PR #26, opened 2015-02-24
Referral handler

dfish3r:
Items of interest:
* Request#getReferralHandler() returns the raw type
* Default ReferralConnectionFactory is an inner class on AbstractReferralHandler
* SearchReferralHandler leverages an inner class to provide an implementation of SearchReferenceHandler

### DIFF
--- a/apache-provider/src/main/java/org/ldaptive/provider/apache/ApacheLdapConnection.java
+++ b/apache-provider/src/main/java/org/ldaptive/provider/apache/ApacheLdapConnection.java
@@ -145,11 +145,6 @@ public class ApacheLdapConnection implements ProviderConnection
   public Response<Void> bind(final BindRequest request)
     throws LdapException
   {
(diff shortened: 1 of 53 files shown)


serac:
I reviewed as best I could. I understand the design and it looks sound generally. From a broader perspective I'm not sure I agree the complexity is worth the value of referral handling in general. While I'm sure it works for the test cases you had, I think it's fair to wonder whether the variables and complexities of real-world environments will create a morass of feature creep and bug fixes.

One design aspect I noted was the need for casting to the raw type of ReferralHandler in some cases since the parameter types are unknown in those contexts. It's admittedly a little ugly, but it works and I don't have a better suggestion offhand.

dfish3r:
Since this patch was initiated due to security concerns related to providers exposing credentials in the clear, I think it's the best compromise I could arrive at.
At least it addresses that in particular.
Avoid referrals will still be the party line.

Issue #29, opened 2015-03-12
Add resolvedDN to the authentication response


dfish3r:
The AuthenticationResponse does not include the DN that was resolved for authentication.
This DN may or may not be the same as the DN returned on the LdapEntry.

Update the AuthenticationResponse to include the resolved DN.

PR #30, opened 2015-03-17
Add #resolvedDn property to auth response.

dfish3r:
See #29.


serac:
Looks like a useful feature.

PR #38, opened 2015-05-11
Add DnParser#substring(begin, end).

dfish3r:


core/src/main/java/org/ldaptive/DnParser.java
```diff
@@ -75,26 +75,67 @@ public static String getValue(final String dn, final String name)
...
+    }
+
+    final List<LdapAttribute> attrs = convertDnToAttributes(dn);
+    if (endIndex > attrs.size()) {
+      throw new IndexOutOfBoundsException("endIndex cannot be larger than the number of RDNs");
+    }
+
+    final StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < attrs.size(); i++) {
+      if (i >= beginIndex && i < endIndex) {
```
serac: Shouldn't the upper bound be `i <= endIndex`?

dfish3r: Isn't that exactly the semantics that are enforced?
If not, provided a unit test to demonstrate.

serac: You need to clarify the bounds on the parameters. I just checked String#substring, which works exactly like what you have done (including naming), so there's a solid precedent. You just need better javadocs on the acceptable parameter values for indices.

dfish3r: Updated javadocs.

serac:
Usage is clear now, thanks.

PR #57, opened 2015-11-13
add FreeIPA support

tduehr:
Adds support for FreeIPA.

### DIFF
--- a/core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
+++ b/core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
@@ -1,0 +1,105 @@
+/* See LICENSE for licensing and NOTICE for copyright. */
+package org.ldaptive.auth.ext;
+
+import java.util.Calendar;
+import org.ldaptive.LdapAttribute;
+import org.ldaptive.LdapEntry;
+import static org.ldaptive.ResultCode.*;
+import org.ldaptive.auth.AuthenticationResponse;
+import org.ldaptive.auth.AuthenticationResponseHandler;
+import org.ldaptive.io.GeneralizedTimeValueTranscoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Attempts to parse the authentication response and set the account state using data associated with eDirectory. The
+ * {@link org.ldaptive.auth.Authenticator} should be configured to return 'passwordExpirationTime' and
+ * 'loginGraceRemaining' attributes so they can be consumed by this handler.
+ *
+ * @author  tduehr
+ */
+public class FreeIPAAuthenticationResponseHandler implements AuthenticationResponseHandler
+{
+
+  /** Maximum password age. */
+  private int maxPasswordAge = -1;
+
+  /** Maximum password age. */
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
```diff
@@ -0,0 +1,105 @@
...
+    if (response.getResultCode() != SUCCESS) {
+      final FreeIPAAccountState.Error fError = FreeIPAAccountState.Error.parse(response.getResultCode(), response.getMessage());
+      if (fError != null) {
+        response.setAccountState(new FreeIPAAccountState(fError));
+      }
+    } else if (response.getResult()) {
+      final LdapEntry entry = response.getLdapEntry();
+      final LdapAttribute expTime = entry.getAttribute("krbPasswordExpiration");
+      logger.info("krbPasswordExpiration: {}", expTime);
+      final LdapAttribute loginRemaining = entry.getAttribute("loginGraceRemaining");
```
dfish3r: I can't find any documentation indicating that FreeIPA uses the loginGraceRemaining attribute. Was this a copy paste vestige?

core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
```diff
@@ -0,0 +1,105 @@
...
+ * @author  tduehr
+ */
+public class FreeIPAAuthenticationResponseHandler implements AuthenticationResponseHandler
+{
+
+  /** Maximum password age. */
+  private int maxPasswordAge = -1;
+
+  /** Maximum password age. */
+  private int maxLoginFailures = -1;
```
dfish3r: There's no way to set this variable. Looks like the variable is needed, I assume this is an oversight.

tduehr: Correct.

core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
tduehr: Yes, sort of. I started with EDirectory and the beginnings of a Kerberos/LDAP AccountState based on the LDAP password management standard. I'll take a look through the code and send another PR to correct these.

dfish3r: Ok. Make sure you review latest rev on master. Thanks.

tduehr: Seems you guys already took care of this. Thanks.

dfish3r: Can you confirm the code in master is working correctly for your FreeIPA instance. We can't test against FreeIPA. Thanks.

tduehr: I have these files in my CAS 4.1 maven overlay and can confirm they work.

Issue #60, opened 2015-12-01
JDK8: Add support for new date classes


dfish3r:
Java 8 has a new date API.
Deprecate calendar in favor of the new API.

dfish3r:
Fixed in

https://github.com/vt-middleware/ldaptive/commit/74002f5e803487517fcb02e5da738fa2c16c0089

PR #62, opened 2015-12-18
reorg parsers, add AuthenticationResponseHandler parsers, add Sasl configs

tduehr:
Moved parsers to a package below the namespace handler.

Added namespace for AuthenticationResponseHandlers and Sasl bind configurations.


tduehr:
Added to the existing parser tests.

tduehr:
Should be ready to go. Still have to test using my CAS install.

dfish3r:
I'll try to review this week.
The java 8 time push will create some conflicts, but I should be able to navigate.
You may want review: https://github.com/vt-middleware/ldaptive/issues/60

tduehr:
I'll take a look.

tduehr:
This still needs some tweaks anyway...

tduehr:
Now it's ready. (and rebased against master)

dfish3r:
You added copy constructors to classes in org.ldaptive.sasl, but I don't see them used in your changes.
Did you have a use case for those outside of this patch?

tduehr:
They may have been for a previous version of the changes before I realized only two of the SaslConfig child classes have a realm field. It's likely they can be safely removed.

dfish3r:
Cleaned up this pull request a bit, tracking changes on this branch:
https://github.com/vt-middleware/ldaptive/tree/reorg-parsers

dfish3r:
Summary of changes from your patch:
* Uses a single element 'sasl-bind-search-authenticator' instead of distinct elements for each mechanism. i.e. gssapi-search-authenticator, crammd5-search-authenticator, digestmd5-search-authenticator, etc
* Allows any sasl-bind-search-authenticator to set a realm even though not every mechanism uses a realm
* Defines password policy elements as child elements of the authenticator and removed the usePasswordPolicy attribute

tduehr:
Pulling in those changes and cleaning up for checkstyle.

dfish3r:
You don't need to do that work. If you just +1 the branch I referenced I can merge it and close this pull request.

tduehr:
Aren't PasswordPolicyAuthenticationResponseHandler and PasswordExpirationResponse handler normally used together?

dfish3r:
No. I only wired them up together so I could use a single boolean property. They represent entirely different RFCs.

tduehr:
In CAS at least they're used together.

http://jasig.github.io/cas/4.1.x/installation/LDAP-Authentication.html

dfish3r:
They probably did the same thing I did, which is to support a small set of password policy. Once you move into supporting each handler specifically and not just 'password policy', you may as well break things out.

The technical answer to your question is that there is no harm in sending both type of controls as long as they are not critical. The directory should just ignore any control it doesn't support. There may be directories that support both of these controls, but I can't imagine someone implementing a policy that leverages both of them at the same time.

tduehr:
Right, you might want to add `maxOccurs="unbounded"` to the `authenticationResponseHandlerType` element in the schema then.

dfish3r:
I can't imagine that use case. If one shows up I'm happy to change it, but from where I sit a deployer should only ever configure a single password policy handler.

tduehr:
@mmoayyed do you know why CAS documentation shows the use of `PasswordPolicyAuthenticationResponseHandler` and `PasswordExpirationResponse` for LPPE when only one should be needed?

mmoayyed:
Were you referring to 4.2 docs?

mmoayyed:
Ah, figured out the docs. I "suspect" at some point that one handler took of controls that the other did not, but it's also entirely possible that docs are inaccurate, or too flexible.

mmoayyed:
let me ask this: are there controls in one handler that cannot be handled via another? I would assume that `PasswordExpirationAuthenticationResponseHandler` handles password expiration only, but I don't know exactly what `PasswordPolicyAuthenticationResponseHandler` does. That name sounds too generic to me, so I think it's a catch-all component that handles the most generic set of controls. 

What are the other types of handler one can configure? (password policy related or not, of course)

tduehr:
None that make sense to be configured concurrently as it's unlikely someone would have an environment with multiple directory servers as potential authentication backends.

Another solution there could just be to add a delegating response handler that can be configured to contain any combination of handlers needed.

dfish3r:
`PasswordPolicyAuthenticationResponseHandler` should be used with directories that support this:
http://tools.ietf.org/html/draft-behera-ldap-password-policy-10

`PasswordExpirationAuthenticationResponseHandler` should be used with directories that support this:
http://tools.ietf.org/html/draft-vchu-ldap-pwd-policy-00

mmoayyed:
> unlikely someone would have an environment with multiple directory servers as potential authentication backends.

This isn't unlikely at all. It happens every day. But even if multiple backends are available, since policies are tied to authenticators, there could be separate authenticators for each backend. So moot really. 

Thanks for the notes @dfish3r . I'll proceed to update the docs on the CAS side.

dfish3r:
@tduehr and @mmoayyed can you confirm the reorg-parsers branch works for you?

mmoayyed:
Sure. Stand by please.

tduehr:
I'll have to redo some config. I'll give it a shot tonight.

mmoayyed:
One issue, which is sort of irrelevant to this pull:

```
[INFO] LDAPTIVE PARENT .................................... SUCCESS [01:27 min]
[INFO] LDAPTIVE CORE ...................................... SUCCESS [02:31 min]
[INFO] LDAPTIVE BEANS ..................................... SUCCESS [ 13.994 s]
[INFO] LDAPTIVE JSON ...................................... SUCCESS [ 11.073 s]
[INFO] LDAPTIVE TEMPLATES ................................. SUCCESS [  7.434 s]
[INFO] LDAPTIVE WEB APPLICATION ........................... SUCCESS [  1.716 s]
[INFO] ------------------------------------------------------------------------
```

Parent pom does not build the unboundid module.

mmoayyed:
None of the other providers in fact. Is this done on purpose?

mmoayyed:
I may have ask this before, sorry if it's a duplicate, but why doesn't the AD authenticator have the ability to get a policy associated with it? Is that handled automatically and requires no such config? 

That said, things work fine.

mmoayyed:
Few more things, since I noticed there is now support for connection config and search executors via parsers:

1. connection-config seems to be missing a few key attributes, like the initializer DN and password
2. can the ldapUrl attribute be optional? For use cases where it needs to be dynamically set.
3. Can the ldapUrl of the connection pool be optional? Same reason as above.

Something like this:

```
    <ldaptive:connection-config connectTimeout="${ldap.connectTimeout}"
                                id="provisioningConnectionConfig"
                                managerDN=""
                                password=""
                                useStartTLS="${ldap.useStartTLS:false}" />

    <ldaptive:connection-pool blockWaitTime="${ldap.pool.blockWaitTime}"
                              prunePeriod="${ldap.pool.prunePeriod}"
                              idleTime="${ldap.pool.idleTime}"
                              maxPoolSize="${ldap.pool.maxSize}"
                              minPoolSize="${ldap.pool.minSize}"
                              validateOnCheckOut="${ldap.pool.validateOnCheckout}"
                              validatePeriodically="${ldap.pool.validatePeriodically}"
                              validatePeriod="${ldap.pool.validatePeriod}" />
```

dfish3r:
> Parent pom does not build the unboundid module.

All the providers (and some other modules) are controlled via profiles. There is a script in the base directory called `mvn_cmd` which can simplify maven build commands. Otherwise, just activate the profile with -Punboundid.

dfish3r:
> why doesn't the AD authenticator have the ability to get a policy associated with it?

I wire in the ActiveDirectoryAuthenticationResponseHandler by default. It's worth noting that deployers won't be able to set the properties on this class. That's intentional because those properties require elevated access to AD to read the required attributes. Given the complexity of that setup, I decided deployers should just break out and use standard spring beans if they need that functionality.

mmoayyed:
Thanks. All is fair. Let me know what you think about my 1-2-3 above. That list would very much simplify things for me.

dfish3r:
> Let me know what you think about my 1-2-3 above. 

1 - confirmed
2 - Sure, sort of begs the question if any attribute should be required. If ldapURL isn't known until runtime, most other properties can fall into the same category
3 - Same as above

mmoayyed:
On 2, true. You may want to consider most if not all attributes as optional, and if there is enough of them provided, validate the configuration. Otherwise, leave to the runtime.

dfish3r:
Take another look at that branch. I pushed an attempted fix.

mmoayyed:
Excellent. I'll try tomorrow. Stand by please.

mmoayyed:
Looking very good. They only thing I think I have found is that while creating connection pools, the ldapUrl attribute is indeed optional, but it is still required when Spring reads the context. The IDE is fine with it not being there but on startup, I get errors like:

```java
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name '(inner bean)#d675f9f': Error setting property values; nested exception is org.springframework.beans.PropertyBatchUpdateException; nested PropertyAccessExceptions (1) are:
PropertyAccessException 1: org.springframework.beans.MethodInvocationException: Property 'ldapUrl' threw exception; nested exception is java.lang.IllegalArgumentException: Input cannot be empty
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyPropertyValues(AbstractAutowireCapableBeanFactory.java:1518)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1226)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:543)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:482)
	at org.springframework.beans.factory.support.BeanDefinitionValueResolver.resolveInnerBean(BeanDefinitionValueResolver.java:299)
	... 53 more
Caused by: org.springframework.beans.PropertyBatchUpdateException; nested PropertyAccessExceptions (1) are:
PropertyAccessException 1: org.springframework.beans.MethodInvocationException: Property 'ldapUrl' threw exception; nested exception is java.lang.IllegalArgumentException: Input cannot be empty
	at org.springframework.beans.AbstractPropertyAccessor.setPropertyValues(AbstractPropertyAccessor.java:121)
	at org.springframework.beans.AbstractPropertyAccessor.setPropertyValues(AbstractPropertyAccessor.java:75)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyPropertyValues(AbstractAutowireCapableBeanFactory.java:1514)
	... 57 more
```

dfish3r:
Try again, I pushed some more fixes.

mmoayyed:
Much better. All systems go.

Issue #64, opened 2016-01-04
Support passing a context into the authentication workflow


dfish3r:
Authentication could be enhanced by supporting an arbitrary context that could inform both DN resolution and binding.

The specific use case for this feature is supporting velocity templates for creating the search filter in DN resolution.

dfish3r:
Merge request accepted.

PR #65, opened 2016-01-04
Add authentication request context.

dfish3r:
Add new User object to encapsulate user id and context.
Modify DnResolver interface to use User.
See #64.


serac:
Changes look fine to me, but I would be more enthusiastic if I understood the use case better. What sort of context object attached to the user would be helpful for building search filters via Velocity or another template framework?

dfish3r:
Added an integration test to demonstrate.

serac:
It took a bit to put the pieces together mentally, but that does help clarify the use case.

PR #66, opened 2016-01-07
add debugging print for multiple Dn condition

tduehr:


dfish3r:
Fixed in v2 at 78b178ac7a64dcc7ffc3923e3aedf0605cff5fc7

PR #69, opened 2016-01-28
Added DNS SRV strategy to ConnectionStrategy

cperrin88:


dfish3r:
I didn't add a static instance of this class because the implementation is not stateless.

cperrin88:
What does that mean in particular?

dfish3r:
The class has mutators which if invoked could break other threads using the same instance.
See DnsSrvConnectionStrategy#setJndiEnvironment
So it's not a good candidate for sharing among various LdapConnections.

dfish3r:
v2 now implements stateful connection strategies.

PR #70, opened 2016-02-03
Backport of #64 into v1.1

dfish3r:
DnResolver now accepts an Object instead of a String.


dfish3r:
Wanted some additional eyeballs on this.
Backported the changed from issue #64, the end result being changing String to Object in order to support the functionality and remain API compatible.

serac:
Looks good. String->Object widening in particular is no big deal.

PR #73, opened 2016-02-22
minor logic correction

tduehr:
no warning if no bad logins have occurred


dfish3r:
So you only want a warning if krbLoginFailedCount > 0. Seems reasonable.

tduehr:
Yes, if `krbLoginFailedCount > 0` there's something to warn about. Otherwise there isn't. It might be better to expose a failed login count but that may not apply to all auth methods.

Issue #58, opened 2015-11-25
Feature request to support Durations


tomzeller:
Hi, how about a feature request for ldaptive to support Durations ?

"blockWaitTime in IdPv3 ldap.properties"
http://shibboleth.net/pipermail/users/2015-November/025164.html

"Incorrect idp.pool.LDAP.validatePeriod property format"
http://shibboleth.net/pipermail/users/2015-November/025528.html

dfish3r:
The Duration annotation lives in the java-support project. That would give ldaptive a dependency on an external library that doesn't exist in maven central.

Is there another annotation, perhaps in the javase, that could be used?

mmoayyed:
Would @Size work? Most Durations I know of are concrete classes, and live in the javax caching API or in JDK8. JDK8 probably is the best option, but that does require, well, JDK8 :)

dfish3r:
I'm planning to move to JDK8 very soon.

mmoayyed:
Good on you :) Possible timeline?

dfish3r:
Resolved in merge #77

tomzeller:
Nice !

I don't think there is a related issue for the Shibboleth IdP to incorporate this, is that something you can handle too Daniel ?

PR #77, opened 2016-03-04
Add support for java.time.Duration.

dfish3r:
See #58.

Conversion of long and int properties to Duration.
Note that method names didn't change, so PoolConfig#setValidatePeriod now accepts a Duration.
Unfortunately Spring does not yet support Duration with it's own converter, so I have to provide my own in spring beans XML.


core/src/main/java/org/ldaptive/SearchRequest.java
```diff
@@ -29,7 +30,7 @@
   private SearchScope searchScope = SearchScope.SUBTREE;
 
   /** Time search operation will block. */
-  private long timeLimit;
+  private Duration timeLimit = Duration.ZERO;
```
serac: Is there a reason this default (ZERO) is different from those in ConnectionConfig (null)? Seems you'd want consistent defaults in all cases.

integration/src/test/java/org/ldaptive/PropertiesTest.java
```diff
@@ -96,7 +97,7 @@ public void parserProperties(final String bindDn, final String host)
 
     AssertJUnit.assertEquals(host, cc.getLdapUrl());
     AssertJUnit.assertEquals(bindDn, ci.getBindDn());
-    AssertJUnit.assertEquals(8000, cc.getConnectTimeout());
+    AssertJUnit.assertEquals(Duration.ofSeconds(8), cc.getConnectTimeout());
```
serac: So much more readable.

integration/src/test/resources/ldap_jaas.config
```diff
@@ -225,13 +225,13 @@ ldaptive-props {
     ldapUrl="${ldapTestHost}"
     bindDn="${ldapBindDn}"
     bindCredential="${ldapBindCredential}"
-    connectTimeout="8000"
+    connectTimeout="PT8S"
```
serac: The ISO period/duration syntax may take some users time to learn, but it's very meaningful once you get the hang of it. No more ambiguity about what the units are (s/ms/min).

serac:
This is a great change that will make reading configuration much clearer. My only concern is taking this opportunity to define a consistent Duration value that means "wait forever." Reasonable values are -1, 0, and null. I recommend picking one and using it consistently. I would not rely on provider defaults; pass in whatever the library default in all cases so you get consistent behavior.

core/src/main/java/org/ldaptive/SearchRequest.java
dfish3r: The ConnectionConfig properties simulated a null by using -1, those properties now allow null.
(Providers specific handling of those properties make their use somewhat nebulous.)
timeLimit doesn't allow null; setting it to zero keeps the same default value.
(Since timeLimit is a defined part of a search request, it's meaning is the same across providers.)

core/src/main/java/org/ldaptive/pool/IdlePruneStrategy.java
```diff
@@ -81,8 +82,11 @@ public long getPrunePeriod()
...
    * @param  period  to set
    */
-  public void setPrunePeriod(final long period)
+  public void setPrunePeriod(final Duration period)
   {
+    if (period == null || period.isNegative()) {
+      throw new IllegalArgumentException("Prune period cannot be null or negative");
+    }
     prunePeriod = period;
   }
```
dfish3r: Any comment on method signatures that contain 'Period', but accept Duration?

integration/src/test/resources/spring-jaas-context.xml
```diff
@@ -2,10 +2,17 @@
...
+
+  <bean id="conversionService"
+        class="org.springframework.context.support.ConversionServiceFactoryBean">
+    <property name="converters">
+      <set>
+        <bean class="org.ldaptive.beans.spring.convert.StringToDurationConverter"/>
+        <bean class="org.ldaptive.beans.spring.convert.DurationToStringConverter"/>
+      </set>
+    </property>
+  </bean>
```
dfish3r: Any comment on now having to configure the conversion service in order to use spring beans?

serac: It's a necessary requirement at the moment, but I imagine Spring will catch up so that it won't be needed in the future.

core/src/main/java/org/ldaptive/pool/IdlePruneStrategy.java
serac: It's a little irritating to me, but I imagine many folks won't even notice.

dfish3r: I can make API breaking changes here if we want.

serac: Not necessary. I think "period" is the more natural term for describing the property, so it's just a little unfortunate that the best JSE component is named "Duration."

dfish3r:
Ok. Proceed with the merge.

PR #85, opened 2016-04-08
Modify DnResolver changes for API compatibility.

dfish3r:
Added DnResolverEx interface and User class.
Updated applicable DnResolver implementations to implement both DnResolver and DnResolverEx.

### DIFF
--- a/core/src/main/java/org/ldaptive/auth/AuthenticationRequest.java
+++ b/core/src/main/java/org/ldaptive/auth/AuthenticationRequest.java
@@ -108,7 +108,7 @@ public class AuthenticationRequest
   /**
    * Sets the user.
    *
-   * @param  id  of the user
+   * @param  id  of the user to authenticate
    */
   public void setUser(final String id)
   {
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/ldaptive/auth/AuthenticationRequest.java
```diff
@@ -71,9 +110,20 @@ public Object getUser()
...
+  public void setUser(final String id)
+  {
+    user = new User(id);
+  }
+
+
+  /**
+   * Sets the user.
+   *
+   * @param  u  of the user
```
serac: The parameter javadoc makes no sense. Should be "the user to authenticate".

serac:
I noted that `FormatDnResolver` doesn't implement the new `DnResolverEx` interface. Was that intentional?

dfish3r:
I didn't have another impl that needed the new functionality. But I'll add it for consistency.

core/src/main/java/org/ldaptive/auth/AuthenticationRequest.java
dfish3r: I'll fix.

dfish3r:
Pushed some changes, merge if you're satisfied.

PR #90, opened 2016-04-20
Add Authenticator#returnAttributes.

### LINKED ISSUE #89: Allow configuration of return attributes on the Authenticator
Return attributes are defined as part of the authentication request.
It would be useful if those attributes could be supplemented with additional attributes that applies to all users.
In particular, several password policy implementation require that certain attributes be returned.

Update ```Authenticator``` with a returnAttributes property.

dfish3r:
Add returnAttributes for each authentication response handler.
Update spring parser to set appropriate return attributes per authentication response handler.
Fixes #89.

### DIFF
--- a/core/src/main/java/org/ldaptive/auth/ext/ActiveDirectoryAuthenticationResponseHandler.java
+++ b/core/src/main/java/org/ldaptive/auth/ext/ActiveDirectoryAuthenticationResponseHandler.java
@@ -22,6 +22,8 @@ import org.ldaptive.auth.AuthenticationResponseHandler;
 public class ActiveDirectoryAuthenticationResponseHandler implements AuthenticationResponseHandler
 {
 
+  /** Attributes needed to enforce password policy. */
+  public static final String[] ATTRIBUTES = new String[] {"msDS-UserPasswordExpiryTimeComputed", "pwdLastSet", };
 
   /** Amount of time since a password was set until it will expire. Used if msDS-UserPasswordExpiryTimeComputed cannot
    * be read. */
@@ -115,9 +117,6 @@ public class ActiveDirectoryAuthenticationResponseHandler implements Authenticat
    */
   public void setExpirationPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Expiration cannot be null");
-    }
     expirationPeriod = period;
   }
 
@@ -140,9 +139,6 @@ public class ActiveDirectoryAuthenticationResponseHandler implements Authenticat
    */
   public void setWarningPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Warning cannot be null");
-    }
     warningPeriod = period;
   }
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/ldaptive/auth/ext/ActiveDirectoryAuthenticationResponseHandler.java
```diff
@@ -115,9 +117,6 @@ public Period getExpirationPeriod()
    */
   public void setExpirationPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Expiration cannot be null");
-    }
```
serac: Why did you remove this null check? Seems unrelated to the refactoring described in the issue.

core/src/main/java/org/ldaptive/auth/ext/ActiveDirectoryAuthenticationResponseHandler.java
```diff
@@ -140,9 +139,6 @@ public Period getWarningPeriod()
    */
   public void setWarningPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Warning cannot be null");
-    }
```
serac: Same as above.

core/src/main/java/org/ldaptive/auth/ext/EDirectoryAuthenticationResponseHandler.java
```diff
@@ -89,9 +92,6 @@ public Period getWarningPeriod()
    */
   public void setWarningPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Warning cannot be null");
-    }
```
serac: Same as above.

core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
```diff
@@ -158,9 +165,6 @@ public Period getExpirationPeriod()
    */
   public void setExpirationPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Expiration cannot be null");
-    }
```
serac: Same as above

core/src/main/java/org/ldaptive/auth/ext/FreeIPAAuthenticationResponseHandler.java
```diff
@@ -183,9 +187,6 @@ public Period getWarningPeriod()
    */
   public void setWarningPeriod(final Period period)
   {
-    if (period == null) {
-      throw new IllegalArgumentException("Warning cannot be null");
-    }
```
serac: Same as above.

core/src/main/java/org/ldaptive/auth/ext/ActiveDirectoryAuthenticationResponseHandler.java
dfish3r: The check is an artifact of the conversion to ```Period```, when negatives values were not allowed.
It's no longer correct, these variables can be null.

I foresee a use case where configuration is using the full constructor and allowing nulls to pass through will make for easier wiring.

serac:
Looks good generally. I noted the removal of several null checks, which appears to be an unrelated change but perhaps is justifiable.

dfish3r:
Ok. Merge if you are satisfied.

Issue #99, opened 2016-05-28
ClassNotFoundException in JNDI provider is thrown when custom ssl socket factory is used


floragunncom:
new InitialLdapContext() in JndiConnectionFactory.createInternal can throw a ClassNotFoundException when java.naming.ldap.factory.socket is set and the context classloader of the current thread does not "know" the class. For details refer to http://docs.oracle.com/javase/jndi/tutorial/beyond/misc/classloader.html and http://www.javaworld.com/article/2077344/core-java/find-a-way-out-of-the-classloader-maze.html

PR to fix this: https://github.com/vt-middleware/ldaptive/pull/98

dfish3r:
Under what circumstances have you seen this exception occur?

floragunncom:
Every time the context classloader is one which is not aware of the ThreadLocalTLSSocketFactory.class.
We faced this issue in our elasticsearch plugin where we use ldaptive as a library. As a plugin provider we have no direct control over classloaders. 

For us this is a showstopper. Currently we use a forked version of ldaptive with the PR applied but thats not a long term solution. I'd really like to see the PR merged and released. ldaptive is really a great library but it's been a while since the last release. Thx.

dfish3r:
@floragunncom can you confirm that latest master resolves your issue?

dfish3r:
No response from reporter. Assume #100 fixed it.

PR #100, opened 2016-06-02
ClassNotFoundException when custom ssl socket factory is used.

dfish3r:


serac:
This seems like a straightforward solution to the problem described in #99.

PR #104, opened 2016-06-23
ISSUE-96 Updated underlying AF_UNIX implementation to 2.0.4

ememisya:
-Updated underlying AF_UNIX implementation to 2.0.4 (now under com.kohlschutter.junixsocket)
-Removed support for anything other than system property to specify file socket path.
  * To provide the ldapi:// type URL support requires specific implementation per provider, this may be considered in the future.


pom.xml
```diff
@@ -484,6 +484,7 @@
             </file>
           </newVersion>
           <parameter>
+            <ignoreNonResolvableArtifacts>true</ignoreNonResolvableArtifacts>
```
dfish3r: What artifact could not be resolved?

ememisya: ldapi as it was not previously hosted centrally.

Issue #107, opened 2016-07-05
Add support for resolving credential property values from resources


dfish3r:
In some circumstances it is useful to place a credential in a separate file, instead of directly in configuration.
Add support for specifying a credential of the form file:/path/to

dfish3r:
Merged in 21da522584e3414f0918d41c61782b587bc070a2

dfish3r:
Fixed v1.1 at https://github.com/vt-middleware/ldaptive/commit/9140c8f56ef0ff741a367f198944936b9363e1cd

Fixed v1.0 at https://sourceforge.net/p/ldaptive/code/3155/

PR #108, opened 2016-07-11
Add support for reading credential from a resource.

dfish3r:
Only affects the building of AuthenticationRequest and BindConnectionInitializer.
See #107.


serac:
Looks good.

Issue #111, opened 2016-07-18
Make password attribute configurable


rkorn86:
Instead of hardcoding the `userPassword` attribute in https://github.com/vt-middleware/ldaptive/blob/master/core/src/main/java/org/ldaptive/auth/AbstractCompareAuthenticationHandler.java#L67 it would be better to make it configurable. Our company uses different passwords for different services and allow password reading via ACL for different manager binds. So this is a real scenario.

rkorn86:
@dfish3r: @mmoayyed suggested me to write to you, because i implemented a solution for this issue inside Apereo CAS. Having the fix upstream is much better, but I dont understand the ldaptive code very well. I did not find the exact point how to extend your XSL Schema. The following pull request will show, how far I got. Hopefully helpfull.

rkorn86:
@dfish3r do you have any plan, when 1.2 will be finally released or can say, which fixes are missing for a 1.2 release?

dfish3r:
I'm hopeful I can get it done next week.

dfish3r:
v1.1 fixed at https://github.com/vt-middleware/ldaptive/commit/b7b45d94658a7804393dd940b9386b0bc6c344c7

v1.0 fixed at https://sourceforge.net/p/ldaptive/code/3153/

PR #112, opened 2016-07-18
Make user password attribute configurable

rkorn86:
Trying to make fix for #111.

### DIFF
--- a/core/src/main/java/org/ldaptive/auth/AbstractCompareAuthenticationHandler.java
+++ b/core/src/main/java/org/ldaptive/auth/AbstractCompareAuthenticationHandler.java
@@ -22,9 +22,14 @@ public abstract class AbstractCompareAuthenticationHandler extends AbstractAuthe
   /** Default password scheme. Value is {@value}. */
   protected static final String DEFAULT_SCHEME = "SHA";
 
+  /** Default password attribute. Value is {@value}. */
+  protected static final String DEFAULT_ATTRIBUTE = "userPassword";
+
   /** Password scheme. */
   private String passwordScheme = DEFAULT_SCHEME;
 
+  /** Password attribute. */
+  private String passwordAttribute = DEFAULT_ATTRIBUTE;
 
   /**
    * Returns the password scheme.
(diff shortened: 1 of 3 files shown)


beans/src/main/resources/org/ldaptive/beans/spring/spring-ext.xsd
```diff
@@ -109,6 +109,7 @@
       <xsd:extension base="baseAuthenticatorType">
         <xsd:attribute name="baseDn" type="xsd:string" use="required"/>
         <xsd:attribute name="userFilter" type="xsd:string" use="required"/>
+        <xsd:attribute name="passwordAttribute" type="xsd:string" use="optional"/>
```
dfish3r: Were you just making this change for the sake of completeness?
None of the bean definition parsers currently support the CompareAuthenticationHandler.
It could be added as part of this PR, but it will require the implementation of a new authenticator type in the schema.

rkorn86: I didnt even know, if the change in XSD is needed. There is even no place where you specify how to change the password scheme from SHA to something different, so it was hard to guess, whats to correct place to add this this feature.

dfish3r: That's my point. This custom schema doesn't currently support the CompareAuthenticationHandler. Support can be added, but it will require more than just this change.

So if you have a use case for using the <ldaptive/> schema, it can be added.
If you're content with using standard spring <bean/> schema, the changes in the beans package can be removed.

rkorn86:
I have removed the schema changes and will just PR the CompareAuthenticationHandler changes. Can you merge this?

core/src/main/java/org/ldaptive/auth/AbstractCompareAuthenticationHandler.java
```diff
@@ -48,6 +53,28 @@ public void setPasswordScheme(final String s)
   }
 
 
+  /**
+   * Returns the password attribute.
+   *
+   * @return  password scheme
```
dfish3r: Copy paste typo.

rkorn86: typo fixed and  rebase + squashed commit

PR #114, opened 2016-09-20
Duration seconds

dfish3r:


core/src/main/java/org/ldaptive/cache/LRUCache.java
```diff
@@ -75,7 +75,7 @@ protected boolean removeEldestEntry(final Map.Entry<Q, Item> entry)
         }
       }
     };
-    executor.scheduleAtFixedRate(expire, interval.getSeconds(), interval.getSeconds(), TimeUnit.SECONDS);
+    executor.scheduleAtFixedRate(expire, interval.toMillis(), interval.toMillis(), TimeUnit.MILLISECONDS);
```
serac: Why this change? So you can have a Duration down to ms resolution?

dfish3r: Yes, but mostly for consistency.

Issue #81, opened 2016-03-27
Deadlock in ldaptive + CAS 4


jinowolski:
We have experienced deadlocks in our CAS service configured with ldaptive as LDAP connector. It appears to be related to keeping pool locked while performing connection validation. Deadlocks was present in CAS 4.0.1 and 4.2.0 which corresponds to ldaptive versions 1.0.3 and 1.1.0.

In attached jstack, a lot of threads are waiting for the lock <0x000000070f847368> which is owned by Thread-11 daemon performing connection validation.

**[jstack-20160326-1751.txt](https://github.com/vt-middleware/ldaptive/files/191125/jstack-20160326-1751.txt)** (ldaptive 1.1.0, CAS 4.2.0)

While trying to resolve (or at least hotfix) problem I've noticed several things:

- deadlock is more likely to occur when using BlockingConnectionPool rather than SoftLimitConnectionPool
- deadlock is more likely to occur when maximum pool size is greater
- I've tried to force duration of validation to be shorter by adding `-Dcom.sun.jndi.ldap.read.timeout=1000` to Tomcat parameters, but it didn't help
- before deadlock, number of threads grows gradually over about two hours. When it reaches some level (~200 waiting / ~260 total), the tomcat server hangs out

Currently we have the following settings which appears to be quite stable (deadlock occurs rarely):

- validateOnCheckout = false
- **validatePeriodically = true**
- **validatePeriod = 300 (seconds)**
- minPoolSize = 3
- maxPoolSize = 10 [raising to 100 is likely to reproduce problem]
- connectTimeout = 3000 (milliseconds)
- responseTimeout = 3000 (milliseconds)
- blockWaitTime = 3000 (milliseconds)
- idleTime = 600 (seconds)

Part of the cas deployerConfixContext.xml with ldaptive configuration:

    <bean id="authenticator" class="org.ldaptive.auth.Authenticator"
        c:resolver-ref="dnResolver"
        c:handler-ref="authHandler" />

    <bean id="dnResolver" class="org.ldaptive.auth.PooledSearchDnResolver"
        p:baseDn="${ldap.baseDn}"
        p:subtreeSearch="true"
        p:allowMultipleDns="false"
        p:connectionFactory-ref="searchPooledLdapConnectionFactory"
        p:userFilter="${ldap.authn.searchFilter}" />

    <bean id="searchPooledLdapConnectionFactory"
        class="org.ldaptive.pool.PooledConnectionFactory"
[...]

PR #115, opened 2016-09-23
Provide a timeout to enforce on connection validate.

dfish3r:
Add validateTimeout property to PoolConfig.
Use validateTimeout property in AbstractConnectionPool by spawning a thread and waiting for validate completion.
See #81.

### DIFF
--- a/core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
+++ b/core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
@@ -10,9 +10,12 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


dfish3r:
It's worth mentioning other potential implementations, since I have a few concerns about this one.
Here I've applied a timeout to *each* connection validation.
It may make more sense to apply a timeout to the entire validate process.
That seems to make the impact more obvious to a user, in this case if your entire pool is hanging, then you'll have to wait for each connection.
So the total wait time is a function of validateTimeout and pool size at the time of validation.
On the other hand it's easier to decide how long you should wait on a single connection.

Lastly, you could push this solution into the Validator implementation.
That was my first thought, but our implementation doesn't currently have an executor instance and since the pool obviously does, I went the easier route.
I also didn't want users having to wire up an executor (which has to be closed) for this edge case.

serac:
My instinct says don't design for the worst case. If the entire pool is hanging, it's fair to assume there are connectivity issues and connections may not be restored when the wait time elapses. So the exact amount of wait doesn't matter; it's driven by external forces. Now, you could argue that you might wait a little longer than necessary with your design, but that seems marginal by any measure.

serac:
+1 clear and straightforward

Issue #117, opened 2016-10-28
equals delegating to hashCode


dfish3r:
There are several classes where the equals implementation delegates to hashCode.
This is incorrect and can lead to collisions when objects are placed in Collections.

PR #118, opened 2016-10-28
equals should not delegate to hashCode.

dfish3r:
Refactor broken equals to use property comparison.
See #117

### DIFF
--- a/core/src/main/java/org/ldaptive/ad/control/LazyCommitControl.java
+++ b/core/src/main/java/org/ldaptive/ad/control/LazyCommitControl.java
@@ -40,6 +40,15 @@ public class LazyCommitControl extends AbstractControl implements RequestControl
 
 
   @Override
+  public boolean equals(final Object o)
+  {
+    if (o == this) {
+      return true;
+    }
+    return o instanceof LazyCommitControl && super.equals(o);
+  }
+
+  @Override
   public int hashCode()
   {
     return LdapUtils.computeHashCode(HASH_CODE_SEED, getOID(), getCriticality());
--- a/core/src/main/java/org/ldaptive/intermediate/SyncInfoMessage.java
+++ b/core/src/main/java/org/ldaptive/intermediate/SyncInfoMessage.java
@@ -242,7 +242,9 @@ public class SyncInfoMessage extends AbstractIntermediateResponse
              LdapUtils.areEqual(cookie, v.cookie) &&
              LdapUtils.areEqual(refreshDone, v.refreshDone) &&
              LdapUtils.areEqual(refreshDeletes, v.refreshDeletes) &&
-             LdapUtils.areEqual(entryUuids, v.entryUuids);
+             LdapUtils.areEqual(entryUuids, v.entryUuids) &&
+             LdapUtils.areEqual(getControls(), v.getControls()) &&
+             LdapUtils.areEqual(getMessageId(), v.getMessageId());
     }
     return false;
   }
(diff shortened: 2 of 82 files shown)


core/src/main/java/org/ldaptive/ad/control/LazyCommitControl.java
```diff
@@ -40,6 +40,15 @@ public LazyCommitControl(final boolean critical)
 
 
   @Override
+  public boolean equals(final Object o)
+  {
+    if (o == this) {
```
serac: I don't think this self comparison in the control classes is warranted. I could see maybe if you had a static collection of controls that was used throughout the codebase where equality checking would actually return true, but I don't see that usage pattern anywhere. Feels like a premature optimization.

core/src/main/java/org/ldaptive/intermediate/SyncInfoMessage.java
```diff
@@ -230,10 +230,38 @@ public void decode(final byte[] berValue)
...
+      return true;
+    }
+    if (o instanceof SyncInfoMessage) {
+      final SyncInfoMessage v = (SyncInfoMessage) o;
+      return LdapUtils.areEqual(getOID(), v.getOID()) &&
+             LdapUtils.areEqual(messageType, v.messageType) &&
+             LdapUtils.areEqual(cookie, v.cookie) &&
+             LdapUtils.areEqual(refreshDone, v.refreshDone) &&
+             LdapUtils.areEqual(refreshDeletes, v.refreshDeletes) &&
+             LdapUtils.areEqual(entryUuids, v.entryUuids);
```
serac: The fields here are not the same as hashCode -- missing controls and messageId. Probably should be same.

core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
```diff
@@ -988,7 +988,14 @@ public PooledConnectionStatistics getPooledConnectionStatistics()
     @Override
     public boolean equals(final Object o)
     {
-      return LdapUtils.areEqual(this, o);
+      if (o == this) {
```
serac: I'm not sure equality should be defined for component classes. What's the use case?

dfish3r: Since it's a proxy, I wanted equality to pass through to the underlying object. Note that equality isn't defined on the underlying object.

core/src/main/java/org/ldaptive/intermediate/SyncInfoMessage.java
dfish3r: Fixed.

core/src/main/java/org/ldaptive/ad/control/LazyCommitControl.java
dfish3r: Not sure I follow. The control classes do have properties and may not be equal.

core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
serac: Ok, makes sense then that it's a special case of component.

core/src/main/java/org/ldaptive/ad/control/LazyCommitControl.java
serac: No harm of course, just don't see much value.

PR #129, opened 2017-09-08
Hostname verifier refactor

dfish3r:


integration/src/test/java/org/ldaptive/ssl/TLSSocketFactoryTest.java
```diff
@@ -167,21 +231,162 @@ public void connectTLS(final String url)
   public void connectSSL(final String url)
     throws Exception
   {
-    // with no trusted certificates, connection should fail
-    final ConnectionConfig cc = createSSLConnectionConfig(url);
```
serac: You might consider breaking out the tests in this class into individual test methods so you can drive test failure with annotations:

    @Test(expectedAnnotations = {A.class, B.class, C.class}

This would also address the problem of different providers throwing different exception types on SSL handshake errors.

unboundid-provider/src/main/java/org/ldaptive/provider/unboundid/UnboundIDProvider.java
```diff
@@ -83,6 +110,43 @@ protected LDAPConnectionOptions getDefaultLDAPConnectionOptions(final Connection
...
+      @Override
+      public void verifySSLSocket(
+        final String host,
+        final int port,
+        final SSLSocket sslSocket)
+        throws LDAPException
+      {
+        logger.trace("Verifying SSLSocket {} for host {} with verifier {}", sslSocket, host, verifier);
+        try {
+          sslSocket.getSession().getPeerCertificates();
```
serac: This raised an eyebrow. It's clear you need an `SSLSession` in order to perform verification, but I would expect you to have a local variable that stores the reference you get from this operation. Then it would be perfectly clear why this operation precedes the subsequent hostname verification.

dfish3r: It's the call to getPeerCertificates that throws, not getSession.
I'm not sure if a local variable makes that any clearer.

serac: For your consideration:

        logger.trace("Verifying SSLSocket {} for host {} with verifier {}", sslSocket, host, verifier);
        final SSLSession session = sslSocket.getSession();
        try {  
          // Following call is required to initiate SSL handshake
          session.getPeerCertificates();
          if (!verifier.verify(host, session)) {
            throw new LDAPException(ResultCode.CONNECT_ERROR, "Hostname verification failed for " + host);
          }
        } catch (SSLPeerUnverifiedException e) {
          logger.trace("Skipped hostname verification for SSLSocket {} due to handshake failure", sslSocket, e);
        }

dfish3r: Sure. I think a better question is whether hostname verification should succeed when trust fails. I found the errors confusing in that case, which led to this implementation. In hindsight, it may be clearer to simply throw in that case.

dfish3r:
The primary reason for these changes is that the HostnameVerifierConfig really wasn't suitable in the public API. In cases where hostname verification is done via trust, the names to verify must be passed in. Allowing end users to sets those names in cases where they would not (or could not) be used is confusing at best. So HostnameVerifierConfig was replaced with the CertificateHostnameVerifier interface, which I think provides a better API.

unboundid-provider/src/main/java/org/ldaptive/provider/unboundid/UnboundIDProvider.java
serac: Agree that throwing is better on the principle that you really ought to have an impeccable reason for swallowing exceptions. Your ambivalence about the behavior suggests you do not, so throwing is probably the safest bet.

Issue #136, opened 2018-06-28
Clean up thread-local used in ThreadLocalTLSSocketFactory


dfish3r:
The thread-local that contains the `SslConfig` used by `ThreadLocalTLSSocketFactory` is never removed.
Updated `ThreadLocalTLSSocketFactory` to provide a remove method and leverage that method for clean up.

dfish3r:
PR accepted.

PR #137, opened 2018-08-27
Remove threadlocal after use.

dfish3r:
Update the socket factory API to allow removal of the thread local.
Update jndi connection factory to remove thread locals.
See #136.

### DIFF
--- a/core/src/main/java/org/ldaptive/provider/jndi/JndiConnectionFactory.java
+++ b/core/src/main/java/org/ldaptive/provider/jndi/JndiConnectionFactory.java
@@ -53,7 +53,6 @@ public class JndiConnectionFactory extends AbstractProviderConnectionFactory<Jnd
     if (ThreadLocalTLSSocketFactory.class.getName().equals(environment.get(JndiProvider.SOCKET_FACTORY))) {
       final ThreadLocalTLSSocketFactory sf = new ThreadLocalTLSSocketFactory();
       threadLocalSslConfig = sf.getSslConfig();
-      sf.removeSslConfig();
     }
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/ldaptive/provider/jndi/JndiConnectionFactory.java
```diff
@@ -51,7 +51,9 @@ public JndiConnectionFactory(
     environment = Collections.unmodifiableMap(env);
     classLoader = cl;
     if (ThreadLocalTLSSocketFactory.class.getName().equals(environment.get(JndiProvider.SOCKET_FACTORY))) {
-      threadLocalSslConfig = new ThreadLocalTLSSocketFactory().getSslConfig();
+      final ThreadLocalTLSSocketFactory sf = new ThreadLocalTLSSocketFactory();
+      threadLocalSslConfig = sf.getSslConfig();
+      sf.removeSslConfig();
```
serac: The appearance here in the constructor raised a red flag for me, but after having studied the surrounding code I believe I understand that it could be considered correct. The consequence of clearing the thread local here implies there are no other consumers of it, which I assume is the nominal case, but it does beg the question of how it gets set in the first place. I would think it would be safer to use the finalizer pattern to clean up these kinds of resources; is that an option?

dfish3r: See https://github.com/vt-middleware/ldaptive/blob/master/core/src/main/java/org/ldaptive/provider/jndi/JndiProvider.java#L176 for a gnarly example of how this works.

dfish3r: So I suppose the thread local could be removed in a finally block in `JndiProvider`, rather than in the `JndiConnectionFactory` constructor internals.

serac: The component that sets up thread locals should be responsible for cleaning them up, and your suggestion to put the cleanup code in `JndiProvider` is consistent with that advice.

dfish3r: Pushed a fix for this.

Issue #138, opened 2018-09-04
Escape control characters in auth logs


dfish3r:
Active Directory can return non-printable characters in the LDAP response message.
Since this data is often logged, it creates the side effect of causing log parsers to believe the log file is binary.
Escape control characters in the LDAP response message.

dfish3r:
Patch allows the setting of 'org.ldaptive.response.encodeCntrlChars' system property to enable this functionality. Encoding does not happen by default.

PR #139, opened 2018-09-04
Encode cntrl chars in response message.

dfish3r:
Add property to control percent encoding of the message in Response#toString.
See #138

### DIFF
--- a/core/src/main/java/org/ldaptive/LdapUtils.java
+++ b/core/src/main/java/org/ldaptive/LdapUtils.java
@@ -13,6 +13,7 @@ import java.util.Base64;
 import java.util.Collection;
 import java.util.List;
 import java.util.Queue;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.ldaptive.io.Hex;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/ldaptive/LdapUtils.java
```diff
@@ -195,6 +199,37 @@ public static String percentEncode(final String value)
...
+   *
+   * @return  string with percent encoded hex characters
+   */
+  public static String percentEncodeControlChars(final String value)
+  {
+    if (value != null) {
+      final Matcher m = CNTRL_PATTERN.matcher(value);
+      if (m.find()) {
+        final StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < value.length(); i++) {
```
serac: You should be able to implement this more compactly and efficiently using `Matcher#appendReplacement` and `Matcher#appendTail`. See the following javadoc for an example:

https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#appendReplacement-java.lang.StringBuffer-java.lang.String-

In particular, you shouldn't have to explicitly make assumptions about the specific characters in the class that matched like on line 218.

serac:
Looks good generally. I had a minor suggestion for improving some of the replacement logic.

core/src/main/java/org/ldaptive/LdapUtils.java
serac: I believe this is equivalent:

      final StringBuffer sb = new StringBuffer();
      while (m.find()) {
        // Each control character is one byte
        final byte[] bytes = new byte[1];
        // CheckStyle:MagicNumber OFF
        bytes[0] = (byte) (m.group().charAt(0) & 0x7F);
        // CheckStyle:MagicNumber ON
        m.appendReplacement(sb, new String(hexEncode(bytes)));
      }
      m.appendTail(sb);

dfish3r: I explored that solution, but since it creates a new StringBuffer for *every* toString invocation I went a different way. I only wanted to pay the price if a control character was present.

serac: Ok, just wanted to make note of it.

Issue #140, opened 2018-09-17
Passivator orphans connections


dfish3r:
A connection that fails to passivate will not be made available in the pool, but it also won't be closed.

PR #141, opened 2018-09-17
Tear down connection if not valid when returned.

dfish3r:
See #140.

### DIFF
--- a/core/src/main/java/org/ldaptive/pool/BlockingConnectionPool.java
+++ b/core/src/main/java/org/ldaptive/pool/BlockingConnectionPool.java
@@ -167,7 +167,7 @@ public class BlockingConnectionPool extends AbstractConnectionPool
    */
   protected PooledConnectionProxy retrieveAvailableConnection()
   {
-    PooledConnectionProxy pc = null;
+    PooledConnectionProxy pc;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/ldaptive/pool/BlockingConnectionPool.java
```diff
@@ -235,13 +235,13 @@ public void putConnection(final Connection c)
...
     poolLock.lock();
     try {
-      if (active.remove(pc)) {
-        if (valid) {
-          available.add(pc);
-          pc.getPooledConnectionStatistics().addAvailableStat();
-          logger.trace("returned active connection: {}", pc);
-          poolNotEmpty.signal();
-        }
+      if (!valid) {
```
serac: I would think you'd want to remove the connection from the pool if the validity check failed implicitly (throws) or explicitly (returns false). It appears that the `Validator` and `Passivator` components that ship with _ldaptive_ do the right thing and return false in the face of exceptions, but the component interface doesn't really speak to what happens when errors occur. Given the importance of the validity check in maintaining pool state, you might want to code defensively here and add some exception handling either around the `validateAndPassivateConnection` call or internally to the method.

dfish3r: Pushed a fix to add try-catch blocks.

PR #149, opened 2019-08-24
Refactor connection strategies to be stateful.

dfish3r:
Add general purpose RetryMetadata class to facilitate backoff algorithms.


dfish3r:
This is the feature you asked for, mark connection URLs inactive when they fail to minimize thrashing for the next connection open.

The ugliness here is the locking (+ synchronization) and the fact that connections now have an additional thread associated with them.
The thread is what tests the inactive connections to make them active in the future.
Note that inactive URLs are simply tried last, after active URLs.

Also note the default inactive period is one minute and default test condition does not to any backoff.

dfish3r:
Pushed changes to use a singleton for URL activation.

dfish3r:
@serac if this looks good, go ahead and merge.

PR #156, opened 2019-09-10
Refactor LdapURLSet and ConnectionStrategy pattern.

serac:
Make the `LdapURLSet` responsible for tracking available URLs with `ConnectionStrategy` adapting to answer a simple question:

Given an `LdapURLSet`, what is the next active URL that should be used for an arbitrary consumer function.

The common consumer function opens a connection to the given LDAP URL. State tracking moves to `LdapURLSet` and the `ConnectionFactory` is responsible for producing and managing exactly one instance of this class; the `ConnectionFactory` instance is passed around to provide access to both the `ConnectionConfig` and the `LdapURLSet` to components that need it.

A sequence diagram in the docs directory describes the new connection initialization process.


serac:
I forgot to pull in your changes that were committed after I branched off for this work. I mostly wanted to get this in front of you to review the change in component concerns. I'm happy to rework it any way you like if you're supportive of the changes.

serac:
Also, I punted on getting the `ConnectionStrategy` tests that do the failover testing to work. I can do that as well if you'd like, but I figured you might be able to do it faster than I could.

dfish3r:
v2 supersedes this issue.

PR #160, opened 2019-11-25
Add support for SCRAM SASL mechanisms.

dfish3r:
Refactor SaslClient to DefaultSaslClient, which applies to code that uses the javax.security.sasl.SaslClient class.
SaslClient is now a generic interface for any SASL impl that needs to perform round trips to the server.
Update the Mechamism enum to support arbitrary String properties.
Don't log SASL credentials in either the request or the response.


dfish3r:
You can largely ignore the API changes in your review, the class that needs your study is `ScramSaslClient`.

PR #163, opened 2019-12-07
Refactor ConnectionTransport classes.

dfish3r:
Rename ConnectionTransport to NettyTransport.
Move thread pool behavior into classes that implement ThreadPoolStrategy.
Provide 3 impls of DefaultThreadPoolStrategy, StatefulThreadPoolStrategy and SingletonThreadPoolStrategy.
System properties can still be used to override the default behavior.


dfish3r:
Probably not exactly what you were looking for, but definitely an improvement.

PR #166, opened 2020-01-09
Remove servlet implementations.

dfish3r:
These implementations existed to support the webapp module, which has been removed.


serac:
We discussed this privately and I was in full agreement these are of no further value.

PR #167, opened 2020-01-24
Refactor LdapException to support result codes.

dfish3r:
This patch changes the LdapException class to require a result code.
While that code may be legitimately null in some cases, this allows the impl to thoughtfully consider the use of client side result codes.

### DIFF
--- a/core/src/main/java/org/ldaptive/transport/netty/NettyConnection.java
+++ b/core/src/main/java/org/ldaptive/transport/netty/NettyConnection.java
@@ -8,7 +8,6 @@ import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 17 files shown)


dfish3r:
@dhawes just wanted a sanity check of the client side codes I chose.

core/src/main/java/org/ldaptive/transport/netty/NettyConnection.java
```diff
@@ -818,17 +821,21 @@ protected void write(final DefaultOperationHandle handle)
                 bindLock.readLock().unlock();
               }
             } else {
-              handle.exception(new LdapException("Bind in progress"));
+              handle.exception(new LdapException(ResultCode.LOCAL_ERROR, "Bind in progress"));
```
dhawes: This one caught my eye because of err=14 (SASL bind in progress), but I think it makes sense in this case after looking at it closer.

dhawes:
I think the codes chosen are fine. I noted one that stood out to me, but it seems correct in context.

+1 to merge from me.

dfish3r:
I believe *any* code >80 is considered a client side error code. (i.e. you'll never get this from the server.) Please indicate which errors you think are server-side. My intention with this PR was to update only client side result codes.

PR #169, opened 2020-02-04
Remove PoolConfig class.

dfish3r:
The PoolConfig class contained some, but not all of the pooling configuration.
Move these properties into the AbstractPool class to provide better consistency for where pooling configuration is stored.


dfish3r:
I believe the `PoolConfig` class hung around for so long for API compatibility.
However, it's reasonable to ask why not remove `AbstractPool` and push its properties into `PoolConfig`.

dfish3r:
I have. `AbstractPool` contains only configuration properties at this point. I think the class only exists to make `AbstractConnectionPool` more readable at this point. It does facilitate the unlikely need for a different connection pool implementation.

You could make the argument that `AbstractPool` should go away and all it's properties should be in `PoolConfig`.

serac:
I think it's a stretch to imagine another kind of resource pool other than `Connection`. Again, reducing the number of types is helpful for folks coming to the library. Configuring an instance of a pool directly makes sense. The only reason you can't do similar on connections is that they are dynamic resources, so the need for `ConnectionConfig` is a special case in that sense.

dfish3r:
Pushed a commit to address your feedback.
Note that refactoring away the config class removed the ability to make some of the pool config immutable.
i.e. you couldn't change the pool size after init. That's possible now.

PR #170, opened 2020-02-16
DERPath performance improvements.

dfish3r:
Remove the ability to register a DERPath by string to emphasize the use of static DERPath objects.
Change regex to use non-capturing group.
Under load the hashCode implementation in DERPath was shown to be a hotspot due to all the collection operations.
Simplify the hashCode implementation to improve performance.

### DIFF
--- a/core/src/main/java/org/ldaptive/asn1/DERPath.java
+++ b/core/src/main/java/org/ldaptive/asn1/DERPath.java
@@ -72,7 +72,7 @@ public class DERPath
     for (UniversalDERTag tag : UniversalDERTag.values()) {
       validNames.append('|').append(tag.name());
     }
-    NODE_PATTERN = Pattern.compile(String.format("(%s)(\\[(\\d+)\\])?", validNames.toString()));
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/ldaptive/asn1/DERPath.java
```diff
@@ -208,7 +208,12 @@ public boolean equals(final Object o)
   @Override
   public int hashCode()
   {
-    return LdapUtils.computeHashCode(HASH_CODE_SEED, nodeStack);
+    int hc = HASH_CODE_SEED;
+    int index = 1;
+    for (Node n : nodeStack) {
+      hc += n.hashCode() * index++;
```
serac: I don't see anything necessarily _wrong_ with this given the impl of `Node#hashCode()`, which should produce a unique value given same nodes in different order. That said, the following is more consistent with the textbook answer:

    hc = HASH_CODE_SEED * hc + n.hashCode()

dfish3r: Pushed a fix.

PR #171, opened 2020-03-05
Add some basic benchmarks.

dfish3r:
Intended to provide some baseline benchmarks to measure throughput performance.
Could provide meaningful, quantitative data for knowing whether a change improved performance.


benchmark/pom.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>ldaptive-benchmark</artifactId>
```
serac: Did you generate this artifact/module using the Maven archtype as recommended in the docs, https://openjdk.java.net/projects/code-tools/jmh/?

dfish3r: I believe I did, but then immediately hacked it up to look like the existing pom files. Did I remove something important?

serac:
I'm not familiar with JMH, but a cursory read of https://www.oracle.com/technical-resources/articles/java/architect-benchmarking.html makes it clear that it's the only way to produce meaningful benchmarks on Java code whose throughput is not governed by IO wait, though it's probably preferable in those cases as well.

benchmark/pom.xml
serac: Not that I can see. I was mostly just curious.

PR #172, opened 2020-03-12
Add maven module to aid in library profiling.

dfish3r:
Configures a docker container that executes an ldaptive class with various configuration parameters.
Exposes the necessary ports so that attaching a profiler is easy.
Supports jconsole and yourkit.

Execute: `mvn install` then `profile/run-compose` to test

### DIFF
--- a/profile/docker-compose.yml
+++ b/profile/docker-compose.yml
@@ -1,0 +1,53 @@
+version: "3.7"
+
+services:
+  ldaptive:
+    image: maven:3.6-jdk-11
+    ports:
+      - 9003:9003
+      - 10001:10001
+    volumes:
+      - $HOME/.m2:/root/.m2
+      - $PWD:/apps/ldaptive
+    environment:
+      - HOST=ldap-test
+      - PORT=389
+      - BASE_DN=ou=test,dc=vt,dc=edu
+      - BIND_DN=uid=1,ou=test,dc=vt,dc=edu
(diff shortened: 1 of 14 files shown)


profile/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+      - BIND_CREDENTIAL=VKSxXwlU7YssGl1foLMH2mGMWkifbODb1djfJ4t2
+      - PROFILE_CLASS
+      - THREAD_COUNT
+      - THREAD_SLEEP
+      - ITERATIONS
+      - USE_YOURKIT
+    networks:
+      ldap_net:
+        ipv4_address: 172.28.1.10
+    command: >
```
serac: I don't see any mechanism to make your client wait on directory startup/initialization.

serac:
Looks good. Just a minor question/comment about docker container startup order.

profile/docker-compose.yml
dfish3r: That's a good point. No wait exists for the integration tests either. Openldap always starts in time....

serac: I think that's a fine answer for test fixtures, but if integrating `wait-for-it.sh` or similar isn't too difficult, then maybe it makes your tools a little more durable.

dfish3r: I think this works because in all my use cases a maven compile happens before tests that depend on Openldap are executed. That gives plenty of time for the Openldap container to start.

serac: Makes sense.

PR #173, opened 2020-03-26
Add basic request throttling implementation.

dfish3r:
The HandleMap is an obvious place to relieve frontside request pressure since it is consulted for all requests and responses.
I'm hesitant to provide any further controls than system properties because I think there will generally be better solutions to limit the pressure from incoming requests.
But in the case where deployers have no other options, the system properties will provide some relief.


serac:
My instinct was to suggest simply using a blocking implementation of Map if throttling is requested such that the capacity of the map determines whether to block. No such collection exists, and by some accounts for good reason. Your solution is clear enough and easy to follow.

PR #175, opened 2020-06-23
Add new ResourceLoader interface.

dfish3r:
Refactor resource loading out of LdapUtils and into new ResourceUtils class.
Provide facility to set custom resource loaders.

### DIFF
--- a/core/src/main/java/org/ldaptive/io/FileResourceLoader.java
+++ b/core/src/main/java/org/ldaptive/io/FileResourceLoader.java
@@ -1,0 +1,38 @@
+/* See LICENSE for licensing and NOTICE for copyright. */
+package org.ldaptive.io;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Creates an {@link InputStream} from a string that is prefixed with 'file:'. See {@link
+ * FileInputStream#FileInputStream(File)}.
+ *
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/ldaptive/io/FileResourceLoader.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Creates an {@link InputStream} from a string that is prefixed with 'file:'. See {@link
+ * FileInputStream#FileInputStream(File)}.
+ *
+ * @author  Middleware Services
+ */
+public class FileResourceLoader implements ResourceLoader
```
serac, line 15: I wonder if we should simply take the opportunity to broaden scope to load anything that `URL#openConnection()` can handle, which would cover the existing filesystem use case as well as the HTTP(S) use case. The component name would be `URLResourceLoader` in that case.

serac:
Looks good!

core/src/main/java/org/ldaptive/io/FileResourceLoader.java
dfish3r: Note that this patch gets us further to that goal with the URLResourceLoader, but does not wire it by default.

serac: Yes, I saw that and don't have concerns at this time.

Issue #176, opened 2020-06-27
NullPointerException when parsing SearchResponse containing Control with Criticality and Value


xweskingx:
*Issue Description*
The AbstractMessage class looks for the following fields when parsing a Control:

```/** DER path to criticality. */
    private static final DERPath CRITICAL_PATH = new DERPath("/SEQ/BOOL");

    /** DER path to OID. */
    private static final DERPath OID_PATH = new DERPath("/SEQ/OCTSTR[0]");

    /** DER path to value. */
    private static final DERPath VALUE_PATH = new DERPath("/SEQ/OCTSTR[1]");
```

Unfortunately, when a Control has both Criticality and Value the Value index will actually be at 2 and not 1 due to the way the DERParser increments the index when parsing.

Sending a SearchResponse containing a Control with Criticality and Value results in the following Exception:

```
io.netty.handler.codec.DecoderException: java.lang.NullPointerException
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:471)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:324)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:296)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:271)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
[...]

PR #177, opened 2020-06-27
#176 - NullPointerException when parsing SearchResponse containing Control with Criticality and Value

xweskingx:
Potential fix for #176 to handle a SearchResponse which has Control with or without criticality (which is optional).

Added unit tests for both cases.

### DIFF
--- a/core/src/main/java/org/ldaptive/AbstractMessage.java
+++ b/core/src/main/java/org/ldaptive/AbstractMessage.java
@@ -223,7 +223,11 @@ public abstract class AbstractMessage implements Message
       parser.registerHandler(CRITICAL_PATH, (p, e) -> critical = BooleanType.decode(e));
       parser.registerHandler(OID_PATH, (p, e) -> oid = OctetStringType.decode(e));
       parser.registerHandler(VALUE_PATH, (p, e) -> value = e.slice());
-      parser.registerHandler(ALT_VALUE_PATH, (p, e) -> value = e.slice());
+      parser.registerHandler(ALT_VALUE_PATH, (p, e) -> {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/ldaptive/AbstractMessage.java
```diff
@@ -220,6 +223,11 @@ public ControlParser()
       parser.registerHandler(CRITICAL_PATH, (p, e) -> critical = BooleanType.decode(e));
       parser.registerHandler(OID_PATH, (p, e) -> oid = OctetStringType.decode(e));
       parser.registerHandler(VALUE_PATH, (p, e) -> value = e.slice());
+      parser.registerHandler(ALT_VALUE_PATH, (p, e) -> {
+        if (value == null) {
+          value = e.slice();
+        }
```
dfish3r, line 229:
```suggestion
        if (value != null) {
          throw new IllegalArgumentException("Multiple control values found");
        }
        value = e.slice();
```

dfish3r:
Looks good. Thanks for this patch. In the review I actually discovered two more bugs related to controls, so this PR was very helpful.

core/src/main/java/org/ldaptive/AbstractMessage.java
dfish3r: Maybe a log message is just as useful here, but some sort of indicator that an invalid sequence was received seems merited.

dfish3r: Actually, this is fine the way it is. It lines up with the way other parsers work.

Issue #179, opened 2020-08-28
support JDK 8 in maven releases


terefang:
hello

i have included the maven dependency in a jdk8 project and getting byte-code version errors on compilation.

please provide a version compatible with jdk8 in maven.

thanks in advance

dfish3r:
The decision to move to java 11 was made based on Oracle's EOL timeline for java 8 at the time. Oracle has since walked back their support deadlines, but we made the best decision we could at the time. 

I'm reviewing what code changes would be needed to support a java8 version, but at this point I'm not sure of the scope of that work.

dfish3r:
Wow, nice work. I started a branch to test out one possibility which keeps everything on the master branch. https://github.com/vt-middleware/ldaptive/pull/181

For now a PR against master is fine. Thank you for contributing.

terefang:
@xweskingx good work

@dfish3r as for the JDK8 EOLness ...

even IBM thinks to keep jdk8/jre8 around at least till 2025. 
see https://www.ibm.com/support/pages/java-sdk-lifecycle-dates

thx

dfish3r:
Latest master can be used to build a Java 8 jar using
`mvn -P-jdk11,jdk8 package`
I also published a -jdk8 jar to the snapshot repository.

Kudos to @xweskingx for providing a patch.

dfish3r:
So I had to give up on using a classifier to support JDK8, it just didn't play nice with the way the artifact bundle was created.
Consequently the coordinates for the artifact have changed.

using this repository:
```
  <repository>
    <id>vt-middleware</id>
    <name>Middleware Open Source Repository</name>
    <url>https://raw.github.com/vt-middleware/maven-repo/master</url>
  </repository>
```

you can declare this dependency:
```
  <dependency>
    <groupId>org.ldaptive</groupId>
    <artifactId>ldaptive-jdk8</artifactId>
    <version>2.0.1-SNAPSHOT</version>
  </dependency>
```

Can you confirm this works for your JDK8 use case? Thanks.

dfish3r:
Ping @terefang @xweskingx

xweskingx:
@dfish3r  Tested with the build from the repo listed above, looks to be working well for my use case. Thanks!

dfish3r:
Thanks @xweskingx. I'll publish the jdk8 artifacts to maven central with the next release.

sysmat:
[...]

PR #180, opened 2020-09-03
#179 - Port library to support jdk8

xweskingx:
Work in Progress.

This contains the source code changes necessary to compile against jdk8, and also disables the test tasks.

Additional changes are necessary to get the tests working.


dfish3r:
Fixed in #181

PR #184, opened 2020-10-16
Remove reference to missing trust manager

### LINKED ISSUE #183: Missing HostnameVerifyingTrustManager per docs
Per the docs:

> If you have a implementation of javax.net.ssl.TrustManager it can be added directory to SslConfig. Ldaptive provides several implementations which may be helpful:
> AggregateTrustManager - combines multiple trust managers allowing the use of all trust managers or any trust manager to pass validation
> DefaultTrustManager - the default JVM trust managers
> HostnameVerifyingTrustManager - custom trust manager based on certificate hostname
> AllowAnyTrustManager - trusts any client or server

`HostnameVerifyingTrustManager` appears to be missing/invalid, and should be removed from the docs possibly.

I can certainly send a pull request to update the docs, if you'd point out where the docs are.

mmoayyed:
Closes #183 

By the way, you shouldn't need to host javadocs in gh-pages branch. It makes the repository size explode after a while. This might be a sufficient option:
https://javadoc.io/doc/org.ldaptive/ldaptive/latest/index.html

### DIFF
--- a/docs/guide/connections.md
+++ b/docs/guide/connections.md
@@ -69,7 +69,6 @@ If you have a implementation of javax.net.ssl.TrustManager it can be added direc
 
 - AggregateTrustManager - combines multiple trust managers allowing the use of all trust managers or any trust manager to pass validation
 - DefaultTrustManager - the default JVM trust managers
-- HostnameVerifyingTrustManager - custom trust manager based on certificate hostname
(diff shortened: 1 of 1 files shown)


dfish3r:
Thanks for the tip on javadoc.io. I was unaware of that service. It seems like a great alternative.

Issue #188, opened 2020-12-21
Connection validator can't block in netty pipeline


dfish3r:
Connection validation is implemented via the context executor on channel activation. That scheduled task should not block the thread pool when it executes.

Refactor connection validation to provide a non-blocking implementation to support this scenario.

dfish3r:
Note that this form of validation is separate from connection pool validation and it's primary use case is with the `SingleConnectionFactory`.

dfish3r:
Fixed in MR #189

PR #189, opened 2020-12-21
Refactor validator to provide async methods

dfish3r:
There are currently 3 distinct use cases for connection validation.
Validating a single connection and waiting for a result.
Validating a single connection without blocking.
Validating multiple connections all at the same time where the operation is async but the results are waited on.

This patch adds two methods to ConnectionValidator called #applyAsync to meet these use cases.
I would have preferred a single method, but this strategy affords a lot of flexibility.
These API changes made the validation code in AbstractConnectionPool much cleaner, eliminating the need for a cached thread pool.
Another ancillary benefit is that the #apply method now enforces the timeout, it no longer needs to be implemented by the invoker.

See #188

### DIFF
--- a/core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
+++ b/core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
@@ -2,9 +2,9 @@
 package org.ldaptive;
 
 import java.time.Duration;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
```diff
@@ -55,6 +64,38 @@ public void setValidateTimeout(final Duration timeout)
...
+      return false;
+    }
+    return applyAsync(conn).get();
+  }
+
+
+  @Override
+  public Supplier<Boolean> applyAsync(final Connection conn)
+  {
+    final BlockingQueue<Boolean> queue = new ArrayBlockingQueue<>(1);
```
serac, line 80: Use of a blocking queue to synchronize on a single boolean value struck me as practical but inefficient, so I took some time to see if I could come up with a more efficient alternative. Sent you a PR for my solution: https://github.com/vt-middleware/ldaptive/pull/190.

core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
```diff
@@ -55,6 +64,38 @@ public void setValidateTimeout(final Duration timeout)
...
+    applyAsync(conn, queue::add);
+    return () -> {
+      Boolean validateResult = null;
+      try {
+        if (Duration.ZERO.equals(getValidateTimeout())) {
+          // this configuration depends on the connection config response timeout for it's wait behavior
```
serac, line 86: ``
```suggestion
          // this configuration depends on the connection config response timeout for its wait behavior
```

core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
```diff
@@ -956,44 +954,19 @@ public void validate()
         logger.debug("validate available pool of size {} for {}", available.size(), this);
 
         final List<PooledConnectionProxy> remove = new ArrayList<>();
-        if (Duration.ZERO.equals(validator.getValidateTimeout())) {
```
serac, line 959: The way this cleaned up is a good sign that this PR is an improvement.

serac (approved):
I think this trades API complexity for added flexibility and simplification of some pretty ugly async handling code in AbstractConnectionPool and a couple other places. I think that's a beneficial tradeoff.

core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
dfish3r: PR merged.

PR #192, opened 2021-02-12
Added a configuration option to the SearchDnResolver to allow returning an attribute in the entry, if present. 

elmendavies:
Dear all,
The configuration parameter proposed and the implementation is to be used in CAS for the following use case. Suppose we have these entries:

Identity entry: 

dn: cn=12345,o=people,dc=example,dc=com
userPassword: 1234
attributeA: AAAA
attributeB: CCCC
attributeC: CCCC

Mail account entry:

dn: mail=username1@company.com,o=accounts,dc=example,dc=com
owner: cn=12345,o=people,dc=example,dc=com
mail: username@company.com

What we want is to allow the user using their mail "username1@company.com" and password "1234" for credentials, without the need to copy and maintain the userPassword field synchronized in both the Identity and Mail account entries. 

This is a proposal: If you think it is a good approach and consider for inclusion in master as is, I will to have time to, first, implement the changes in a modified CAS and, then, check then in a local environment. 

What do you think about the solution?

Thank you very much in advance!

### DIFF
--- a/core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
+++ b/core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
@@ -216,10 +216,7 @@ public class SearchDnResolver extends AbstractSearchOperationFactory implements 
   public void setResolveFromAttribute(final String attributeToUse) {
     logger.trace("setting attributeToUse: {}", attributeToUse);
 
-    if (attributeToUse == null || attributeToUse.isEmpty())
-      this.resolveFromAttribute = null;
-    else
-      this.resolveFromAttribute = attributeToUse;
+    this.resolveFromAttribute = attributeToUse;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


dfish3r:
Interesting use case. At first I thought this may merit a subclass, but there probably isn't enough here for that. In general +1 on the proposal. I'll likely clean up the implementation a bit and possibly rename `resolveDnFromAttribute` to `resolveFromAttribute`.

elmendavies:
I agree. Changed.

I will check if with these changes it is enough to support the use case in CAS (I will implement it to test).

dfish3r:
Any update on this PR?

elmendavies:
I have implemented in CAS in a local branch to test. But I have come with some issues with the build process and I had no time this week to finish. I hope to get time and finish testing in no more than a week. 
BR, Miguel

elmendavies:
Hi Daniel,

I have tested and added the PR in CAS to check whether the approach is right in this way.

https://github.com/elmendavies/cas/tree/use-attribute-to-bind
https://github.com/apereo/cas/pull/5112

During the testing I found that the behavior wasn't correct and I have corrected.

Best regards,
Miguel

core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
```diff
@@ -262,7 +287,21 @@ public String resolve(final User user)
    */
   protected String resolveDn(final LdapEntry entry)
   {
-    return entry.getDn();
+    String dn = entry.getDn();
```
dfish3r, line 290: Configuration for `resolveFromAttribute` should be either/or. That is, if it's turned on it should either produce a DN from an attribute or return null. Not fallback to the entry DN.

core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
```diff
@@ -262,7 +287,21 @@ public String resolve(final User user)
...
+
+    if (resolveFromAttribute != null) {
+      LdapAttribute attr = entry.getAttribute(resolveFromAttribute);
+
+      if (attr.size() != 1) {
+        logger.info("Skipping candidate as it does not meet cardinality (must contain a single value), in dn: " + dn + " resolveDnFromAttribute: " + resolveFromAttribute);
+      } else if(attr.isBinary()) {
+        logger.info("Skipping candidate as it is binary, in dn: " + dn + " resolveDnFromAttribute: " + resolveFromAttribute);
```
dfish3r, line 298: Change these `info` statements to `warn`. Also, use parameterized logging.

core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
```diff
@@ -196,6 +199,28 @@ public void setDerefAliases(final DerefAliases da)
...
+
+  /**
+   * Setsthe attribute to use to resolve the DN. If null, the resolver will take the entry's DN
+   * @param attributeToUse
+   */
+  public void setResolveFromAttribute(final String attributeToUse) {
+    logger.trace("setting attributeToUse: {}", attributeToUse);
+
+    if (attributeToUse == null || attributeToUse.isEmpty())
+      this.resolveFromAttribute = null;
```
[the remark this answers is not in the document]
elmendavies: You are right. I moved the check to cas-server-api-configuration-model:

https://github.com/apereo/cas/pull/5112/commits/2de74c06bc50d4bd2e5f2bb47356d2799091503b

core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
elmendavies: I have rewriten this part and added a new method to make the rule much more explicit.

core/src/main/java/org/ldaptive/auth/SearchDnResolver.java
elmendavies: I agree in both and implemented accordingly.

dfish3r:
This looks like it's ready to go. If so, please remove WIP from the title and I'll merge. Thanks.

elmendavies:
Here you have! Thanks

PR #194, opened 2021-03-24
Allow credential reader to work with multiple paths

### LINKED ISSUE #193: Support multiple trust certificate paths for `X509CredentialConfig`
Given the following code snippet:

```java
var cfg = new X509CredentialConfig();
cfg.setTrustCertificates(getTrustCertificatesAsString());
cc.setSslConfig(new SslConfig(cfg));
```

It appears that, as of ldaptive 2.0.1, `X509CredentialConfig#setTrustCertificates()` generally can only support a single resource when working with a path as a String. This feels odd, specially since the method name `setTrustCertificates()` implies support for multiple certificates should be allowed. I suppose that is still valid if one were to put together a keystore of some kind as a single resource to contain multiple certificates.

For convenience, I want to suggest that `AbstractCredentialReader` may be modified to accept multiple individual paths, separated by comma/space/etc so that multiple trust certificates can be found and set on the credential configuration without the need to combine them in a keystore. 

Something like this:

```java
var cfg = new X509CredentialConfig();
cfg.setTrustCertificates(List.of("cert1.cer", "cert2.cer"));
cc.setSslConfig(new SslConfig(cfg));
```

This feels more intentional and somewhat more obvious, and does require less work from the ldaptive user (no one likes messing about with keystores!) I'd be happy to send a pull request for review, if this is a reasonable enhancement.

mmoayyed:
Closes #193

### DIFF
--- a/core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
+++ b/core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
@@ -1,24 +1,46 @@
 /* See LICENSE for licensing and NOTICE for copyright. */
 package org.ldaptive.ssl;
 
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.security.GeneralSecurityException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 /**
- * Loads X.509 certificate credentials from a classpath, filepath, or stream resource. Supported certificate formats
- * include: PEM, DER, and PKCS7.
+ * Loads X.509 certificate credentials from a classpath, filepath, or stream resource.
+ * When working with filepath, multiple files may be separated using a comma (i.e. {@code cert1.pem,cert2.crt}).
+ * Supported certificate formats include: PEM, DER, and PKCS7.
  *
  * @author  Middleware Services
  */
 public class X509CertificatesCredentialReader extends AbstractCredentialReader<X509Certificate[]>
 {
 
-
+  @Override
+  public X509Certificate[] read(final String path, final String... params)
+      throws IOException, GeneralSecurityException
+  {
+    final String[] paths = path.split(",");
+    return Arrays.stream(paths)
+        .map(individualPath -> {
+          try {
+            return Arrays.asList(super.read(individualPath, params));
+          } catch (final Exception e) {
+            logger.warn(String.format("Unable to read certificate at %s", individualPath), e);
+          }
+          return null;
+        })
+        .filter(Objects::nonNull)
+        .flatMap(List::stream)
+        .toArray(X509Certificate[]::new);
+  }
+  
   @Override
   public X509Certificate[] read(final InputStream is, final String... params)
     throws IOException, GeneralSecurityException
--- a/core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
+++ b/core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
@@ -2,58 +2,51 @@
 package org.ldaptive.ssl;
 
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.FileSystems;
 import java.security.GeneralSecurityException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 
 /**
  * Loads X.509 certificate credentials from a classpath, filepath, or stream resource.
  * When working with filepath, multiple files may be separated using a comma (i.e. {@code cert1.pem,cert2.crt}).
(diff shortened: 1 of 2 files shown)


mmoayyed:
In the test, I had to provide base64-decoded certs before the reader could parse those. Does it make sense to allow the reader to auto-decode a certificate with an option before it can parse and consume it?

core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
```diff
@@ -1,24 +1,46 @@
...
  */
 public class X509CertificatesCredentialReader extends AbstractCredentialReader<X509Certificate[]>
 {
 
-
+  @Override
+  public X509Certificate[] read(final String path, final String... params)
+      throws IOException, GeneralSecurityException
+  {
+    final String[] paths = path.split(",");
```
dfish3r, line 29: I think I prefer that the `path` argument be an array, then the path delimiter is an implementation detail for the invoker.

core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
```diff
@@ -1,24 +1,46 @@
...
+    return Arrays.stream(paths)
+        .map(individualPath -> {
+          try {
+            return Arrays.asList(super.read(individualPath, params));
+          } catch (final Exception e) {
+            logger.warn(String.format("Unable to read certificate at %s", individualPath), e);
```
dfish3r, line 35: I want this exception to bubble out. Especially in the case of reading trust material.

core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
mmoayyed: Sure. Need some help with this. 

If I switch to:

```
public T read(final String[] paths, final String... params) {}
```

That means I'd have to change the parent interface, which means I'd have to remove this method and fix the parent `AbstractCredentialReader` as well, and with that I am not how how to handle the return type. Am I on the right path?

dfish3r: I wasn't going to update the interface or abstract class for this. The ability to parse multiple paths is unique to this one class currently, so I was expecting to add
```
public X509Certificate[] read(final String[] paths, final String... params)
```
directly to `X509CertificatesCredentialReader`

If other implementations need this in the future we can revisit. Thoughts?

dfish3r: If you're leveraging the interface or some other component, I'll need to rethink.

dfish3r: And I can see if you want to use this with `X509CredentialConfig` (which you probably do). This isn't going work. Nevermind.

I think the only thing to change here is maybe use the platform path separator rather than defaulting to comma.

mmoayyed: Thanks for the notes. Sure, will make the change.

core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
mmoayyed: Because this is handled inside a lambda, I could either wrap the exception inside a RuntimeException and throw back or some other unchecked exception, or shy away from using the lambda and instead just use a normal for-loop to collect certificates. What would you recommend?

dfish3r: I would avoid the lambda so you can deliver an `IOException`.

core/src/main/java/org/ldaptive/ssl/X509CertificatesCredentialReader.java
mmoayyed: Sorry to bother again; more questions :)

According to the docs, the path separator character,

> It is a semicolon (;) on Windows and colon (:) on Mac and Unix-based operating systems.

The issue here is that paths as `Resource`s require a classifier/prefix. As in, `file:/path/to.xyz.pem`. This messes up the split function on Mac and Unix. Not sure how to solve this without manually parsing stuff. Would be best to fall back onto comma or space or is there a better way I am missing here?

dfish3r: I would probably use `space`. My second choice would be `comma`. But it's hard to say which would be more problematic. Does spring have a convention?

mmoayyed: Judging by the function in Spring's StringUtils (i.e `commaDelimitedListToStringArray`) and loads of other stuff in that class that deal with CSVs, I think Spring would go for comma most likely. I'd think space would be more problematic on windows, since you'd have paths like "C:\Program Files", and generally it's more accepted on windows to have spaces in paths.

dfish3r: +1 for comma.

Issue #197, opened 2021-05-05
`CompareAuthenticationHandler` skips entry resolution; wrong result code?


mmoayyed:
When using `CompareAuthenticationHandler`, successful authentication attempts return the following:

```java
return new AuthenticationHandlerResponse(compareResponse, compareResponse.isTrue() ? 
   AuthenticationResultCode.AUTHENTICATION_HANDLER_SUCCESS : AuthenticationResultCode.AUTHENTICATION_HANDLER_FAILURE, c);
```

The above response will never be considered success because success for `AuthenticationHandlerResponse` is determined as:

```java
default boolean isSuccess() {
        return ResultCode.SUCCESS == this.getResultCode();
}
```

The result-code for `compareResponse` is always `COMPARE_TRUE`. Since he response is not considered a success, entry resolvers will not run, resulting no returned attributes.

I think `AuthenticationHandlerResponse` should provide an implementation for `isSuccess` that does this?

```java
boolean isSuccess() {
    return super.isSuccess() && getAuthenticationResultCode() == AUTHENTICATION_HANDLER_SUCCESS;
}
```

dfish3r:
Fixed in #198

PR #198, opened 2021-05-05
Adjust success-checking for AuthenticationHandlerResponse

mmoayyed:
Initial proposal for #197


dfish3r:
I think this patch is correct. I'm going to wire up some tests to confirm before I merge it.

PR #202, opened 2021-11-04
Add ValidationExceptionHandler.

dfish3r:
Pooling implementation would benefit from handling validation exceptions on checkout.
Provide an implementation that retries getting a connection.
Bring back ValidationException and ActivationException for general API usage.

### DIFF
--- a/core/src/test/java/org/ldaptive/PooledConnectionFactoryTest.java
+++ b/core/src/test/java/org/ldaptive/PooledConnectionFactoryTest.java
@@ -6,6 +6,8 @@ import java.time.Duration;
 import java.util.Collection;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.util.concurrent.DefaultThreadFactory;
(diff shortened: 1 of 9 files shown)


core/src/test/java/org/ldaptive/PooledConnectionFactoryTest.java
```diff
@@ -252,4 +254,82 @@ public void validatePeriodicallyWithTimeout()
...
+            return Duration.ofSeconds(5);
+          }
+
+          @Override
+          public Boolean apply(final Connection connection)
+          {
+            return validateCount.getAndIncrement() != 1;
+          }
+        })
+        .validateOnCheckOut(true)
```
serac, line 303: My only real concern here is around configuration and arriving at perhaps unexpected behavior in either of the following cases:

1. Setting `validateOnCheckOut(false)` and configuring a `ValidationExceptionHandler`.
2. Setting `validateOnCheckOut(true) and _not_ configuring a `ValidationExceptionHandler`.

The first configuration doesn't do anything, and the second causes an exception to be raised by the pool if the connection is bad. I _think_ the latter behavior is reasonable, but I think a friendlier behavior would be to simply keep trying, which you only get if you configure as you have here in this test case. Thus my biggest concern is the decoupling of these two flags and possibly adding configuration complexity. You might consider simply configuring the retry handler when setting the `validateOnCheckout` flag to true under the assumption that's what most people want; if they have some bizarre requirement to further customize the setter on the factory is still there to do as needed.

serac (commented):
This looks like useful functionality, but I'm concerned it risks adding to configuration complexity that I would characterize as already fairly high for this library. I would recommend that you consider simply making retry the default behavior when `validateOnCheckout` is enabled. I _think_ that's what most people want from a pooling library: always return good connections.

core/src/test/java/org/ldaptive/PooledConnectionFactoryTest.java
dfish3r: I think for (1) the code should emit a warning, I'll make that change. For (2), you're correct, that particular configuration would allow a client to implement their own handling of `ValidationException`.  (Which I'm proposing be the default and you could argue is even currently the default now.)

I'm obviously avoiding any sort of _default_ configuration here because of the potential dangers with retry. The first being, what sort of default configuration would be reasonable? (retry once?) And the second being, retry has the real potential to delay feedback to the caller. In cases where you're already having trouble with connections, asking again (and again) can be much worse than simply returning control with an exception. Particularly when you're dealing with a  server that is slow to connect and/or slow to respond.

dfish3r: I pushed commits to resolve this. I came to terms with using a default retry implementation.

serac: Seems good.

vladimir-mencl-eresearch:
Hi @dfish3r ,

Thanks for your input in the discussion at https://shibboleth.atlassian.net/browse/IDP-1898

I'm trying to assess whether this PR would actually fix the issue I'm facing (described in the above link), where:
* a connection pool is configured against two LDAP servers
* uses the ACTIVE_PASSIVE connection strategy
* and the first server suddenly becomes unresponsive

What I see as desired behaviour is retrying until a connection is established to the other server.

With this implementation (retrying up to `getMaxPoolSize()` times), in the extreme case, when the pool is at the maximum size, with all connections established to the first server, the retries would stop one step short of reaching the other server.

When the pool is not at the maximum size, after using all connections in the pool, there would be an attempt to establish a connection with the other server.

Would you consider changing the number of retries to `getMaxPoolSize() + 1` to accommodate also the situation when the pool is at the maximum size?

Cheers,
Vlad

dfish3r:
One of the features in v2 is auto-reconnect. So in the nominal case, when your active server goes down, connections should attempt a reconnect in the background to the passive server. But let's assume some pesky load balancer that doesn't send TCP resets. In that case you would be dependent on the pool validation to reconnect.

I think you make a good argument for maxSize +1 and an early version of this code worked the same way. So I'll make that change. Note that the code still respects the pool block wait time. So for large pools you may not have enough time to cycle all the connections out of the pool.

vladimir-mencl-eresearch:
Thanks!  Yes, I'm aware the timeout might not still cut it - but this would be a step in the right direction.  Thanks for that!

dfish3r:
Fixed in https://github.com/vt-middleware/ldaptive/commit/e60a233f6e788ed412feabe7f734ab4f106d0469

vladimir-mencl-eresearch:
Thanks!

PR #206, opened 2022-02-18
Add formal package for DN parsing.

dfish3r:
Provide interfaces for the concerns related to parsing DNs.
Improve on the deprecated DnParser class by supporting multi-value RDNs properly.
Position the code to formally support schemas down the road.

### DIFF
--- a/core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
+++ b/core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
@@ -1,0 +1,94 @@
+/* See LICENSE for licensing and NOTICE for copyright. */
+package org.ldaptive.dn;
+
+import org.ldaptive.LdapUtils;
+
+/**
+ * Escapes an attribute value per RFC 4514 section 2.4.
+ *
+ * @author  Middleware Services
+ */
+public class DefaultAttributeValueEscaper implements AttributeValueEscaper
+{
+
+
+  @Override
+  public String escape(final String value)
+  {
+    if (value == null || value.isEmpty()) {
+      return value;
+    }
+
+    final int len = value.length();
+    final StringBuilder sb = new StringBuilder(len);
+    char ch;
+    for (int i = 0; i < len; i++) {
+      ch = value.charAt(i);
+      switch (ch) {
+
+      case '"':
+      case '#':
+      case '+':
+      case ',':
+      case ';':
+      case '<':
+      case '=':
+      case '>':
+      case '\\':
+        sb.append('\\').append(ch);
+        break;
+
+      case ' ':
+        // escape first space and last space
+        if (i == 0 || i + 1 == len) {
+          sb.append('\\').append(ch);
+        } else {
+          sb.append(ch);
+        }
+        break;
+
+      case 0:
+        // escape null
+        sb.append("\\00");
+        break;
+
+      default:
+        // CheckStyle:MagicNumber OFF
+        if (ch > 31 && ch < 127) {
+          sb.append(ch);
+        } else if (i + 1 < len && Character.isHighSurrogate(ch)) {
+          // escape UTF-8 characters
+          final char nextChar = value.charAt(++i);
+          final char[] hex;
+          if (Character.isLowSurrogate(nextChar)) {
+            final int codePoint = Character.toCodePoint(ch, nextChar);
+            hex = LdapUtils.hexEncode(LdapUtils.utf8Encode(new String(new int[] {codePoint}, 0, 1)));
+          } else {
+            hex = LdapUtils.hexEncode(ch);
+          }
+          escapeHex(sb, hex);
+        } else {
+          // escape non-printable ASCII characters
+          escapeHex(sb, LdapUtils.hexEncode(ch));
+        }
+        // CheckStyle:MagicNumber ON
+        break;
+      }
+    }
+    return sb.toString();
+  }
+
+
+  /**
+   * Appends a backslash for every two hex characters.
+   *
+   * @param  sb  to apppend to
+   * @param  hex  to read
+   */
+  private void escapeHex(final StringBuilder sb, final char... hex)
+  {
+    for (int i = 0; i < hex.length; i += 2) {
+      sb.append('\\').append(hex[i]).append(hex[i + 1]);
+    }
+  }
+}
--- a/core/src/main/java/org/ldaptive/dn/Dn.java
+++ b/core/src/main/java/org/ldaptive/dn/Dn.java
@@ -1,0 +1,324 @@
+/* See LICENSE for licensing and NOTICE for copyright. */
+package org.ldaptive.dn;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.ldaptive.LdapUtils;
+
+/**
+ * Distinguished name containing zero or more relative distinguished names. RDNs are ordered from left to right such
+ * that the left-most RDN is considered the first. For the DN 'cn=Jane Doe,ou=People,dc=ldaptive,dc=org', the first RDN
+ * is 'cn=Jane Doe'.
+ *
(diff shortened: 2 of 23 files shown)


core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
```diff
@@ -0,0 +1,94 @@
...
+package org.ldaptive.dn;
+
+import org.ldaptive.LdapUtils;
+
+/**
+ * Escapes an attribute value per RFC 4514 section 2.4.
+ *
+ * @author  Middleware Services
+ */
+public class DefaultAttributeValueEscaper implements AttributeValueEscaper
```
serac, line 11: At first sight this looked out of scope, but I see it's used in `DefaultRDnNormalizer`. I think packaging this functionality in a component is a beneficial change, but putting it in the `dn` package seems wrong. It can obviously stand on its own and is simply an implementation detail of components in this package. Consider moving to a more suitable package.

core/src/main/java/org/ldaptive/dn/Dn.java
```diff
@@ -0,0 +1,324 @@
...
+    rdnComponents.addAll(rdns);
+  }
+
+
+  /**
+   * Returns the first RDN in this DN.
+   *
+   * @return  first RDN
+   */
+  public RDn getRDN()
```
serac, line 86: Use consistent casing.

core/src/main/java/org/ldaptive/dn/Dn.java
```diff
@@ -0,0 +1,324 @@
...
+
+
+  /**
+   * Returns a new DN containing all the RDN components from the supplied index.
+   *
+   * @param  index  of RDNs to include
+   *
+   * @return  DN with sub-components of this DN
+   */
+  public Dn subDN(final int index)
```
serac, line 147: Use consistent casing.

core/src/main/java/org/ldaptive/dn/RDn.java
```diff
@@ -0,0 +1,212 @@
...
+ * Relative distinguished name containing one or more name value pairs. Name value pairs are ordered from left to right
+ * such that the left-most pair is considered the first. For the RDN 'cn=Jane Doe+mail=jdoe@example.com', the first name
+ * value pair is 'cn=Jane Doe'.
+ *
+ * See <a href="http://www.ietf.org/rfc/rfc4514.txt">RFC 4514</a> for more details on the string representations of
+ * RDNs.
+ *
+ * @author  Middleware Services
+ */
+public class RDn
```
serac, line 24: The casing of this class looks really odd. Suggest `RDN` given you already have components with a series of capitals, e.g. `DERParser`.

core/src/main/java/org/ldaptive/dn/RDn.java
```diff
@@ -0,0 +1,212 @@
...
+  public RDn(final String rdn, final DnParser parser)
+  {
+    final List<RDn> rdns = parser.parse(rdn);
+    if (rdns.isEmpty()) {
+      throw new IllegalArgumentException("Invalid RDN: no RDNs found in " + rdn);
+    }
+    if (rdns.size() > 1) {
+      throw new IllegalArgumentException("Invalid RDN: multiple RDNs found in " + rdn);
+    }
+    nameValues = rdns.get(0).getNameValues();
```
serac, line 64: Other constructors appear to protect this set with an immutable wrapper, but I don't see that here. I would recommend delegating to an appropriate constructor.

core/src/main/java/org/ldaptive/dn/RDn.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * Creates a new RDN with a single name value pair.
+   *
+   * @param  attributeName  to add
+   * @param  attributeValue  to add
+   */
+  public RDn(final String attributeName, final String attributeValue)
+  {
+    nameValues = Set.of(new NameValue(attributeName, attributeValue));
```
serac, line 100: I believe `Set.of` returns an immutable list, but I wanted to call it out to double check given behavior of other constructors.

core/src/main/java/org/ldaptive/dn/RDn.java
```diff
@@ -0,0 +1,212 @@
...
+
+
+  /**
+   * Returns a string representation of this RDN, joining each name value pair with '+'.
+   *
+   * @param  normalizer  to apply to the RDN components or null for no formatting
+   *
+   * @return  string form of the RDN
+   */
+  public String format(final RDnNormalizer normalizer)
```
serac, line 158: Perhaps formatting should also be pushed out into a dedicated component for symmetry with parsing. This is the approach of may libraries I've seen, which doesn't preclude convenience methods like this, but it's a good way of separating concerns without compromising convenience. `java.time` date parsing/formatting comes to mind as an example of this pattern.

core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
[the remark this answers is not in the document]
dfish3r: I could go either way on this and I'd probably move it to the base package if anywhere. But an argument for leaving it in the `dn` package is that RFC4514 section 2 is titled: 'Converting DistinguishedName from ASN.1 to a String'.

core/src/main/java/org/ldaptive/dn/RDn.java
dfish3r: I considered that, but there's so little for the formatter to do. The complexity is in the normalizer, which is already externalized.

core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
dfish3r: This is essentially the same comment as above, no?

core/src/main/java/org/ldaptive/dn/Dn.java
dfish3r: Fixed.

core/src/main/java/org/ldaptive/dn/Dn.java
dfish3r: Fixed

core/src/main/java/org/ldaptive/dn/RDn.java
dfish3r: Fixed.

core/src/main/java/org/ldaptive/dn/RDn.java
dfish3r: Correct. `Set.of` is immutable.

core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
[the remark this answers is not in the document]
serac: Ok, we'll leave as-is as a practical concern. I don't think there's anything gained from placement in base package.

core/src/main/java/org/ldaptive/dn/Dn.java
```diff
@@ -0,0 +1,324 @@
...
+ * Distinguished name containing zero or more relative distinguished names. RDNs are ordered from left to right such
+ * that the left-most RDN is considered the first. For the DN 'cn=Jane Doe,ou=People,dc=ldaptive,dc=org', the first RDN
+ * is 'cn=Jane Doe'.
+ *
+ * See <a href="http://www.ietf.org/rfc/rfc4514.txt">RFC 4514</a> for more details on the string representations of
+ * DNs.
+ *
+ * @author  Middleware Services
+ */
+public class Dn
```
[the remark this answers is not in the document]
serac: Ok. Recommend you make a pass at the next convenient time and tighten up that convention. It's a bit jarring to go back and forth.

core/src/main/java/org/ldaptive/dn/DefaultAttributeValueEscaper.java
serac: Yes, same as packaging above.

Issue #203, opened 2022-02-02
Dns Srv record lookup change in behavior between v1.2 and v2


xweskingx:
In version 1.2 of ldaptive, the DefaultConnectionStrategy deferred to jndi for resolving ldap urls.

Using an ldap url such as "ldap://DC=mycompany,DC=org" would use the com.sun.jndi.ldap.ServiceLocator, to convert the ldap url to a dns url "dns:///_ldap._tcp.mycompany.org", and eventually resolve to the DNS Srv record.

After upgrading to version 2 of ldaptive, the default connection strategy is ActivePassiveConnectionStrategy, which does not appear to use the jndi name lookup. 

To address this, it's required to specify the DnsSrvConnectionStrategy when building the connection config.

Unfortunately the DnsSrvConnectionStrategy expects a url in the format "dns://hostname/domain?record" which would look something like "dns:///mycompany.org?_ldap._tcp" for the above example. This requires first converting the ldap url to the dns url manually before building the ldaptive connection config.

It would be nice if the DnsSrvConnectionStrategy would also accept ldap urls of the form "ldap(s):///<ldap dc path>" .

It would also be nice if the default connection strategy would fallback to DnsSrv when ActivePassiveStrategy fails due to no host specified.

This is the behavior of other ldap tools such as the open ldap ldapsearch utility: 

> [ldapsearch] allows you to specify an LDAP URL without the host or port but that uses a base DN that uses only “dc” components (with special characters escaped, like %3D in place of an equal sign and %2C in place of a comma) to indicate that the tool should request the “_ldap._tcp” record for the specified domain (for example, “-H ldap:///dc%3Dexample%2Cdc%3Dcom” will cause it to look up the SRV records for “_ldap._tcp.example.com” and select an appropriate server from that response).

dfish3r:
Thanks for this report. At first glance this feature seems reasonable. I'll review https://datatracker.ietf.org/doc/html/draft-ietf-ldapext-locate-08 and attempt a patch for `DnsSrvConnectionStrategy`.

I'll likely spawn a second issue to evaluate changing the default connection strategy or the behavior of `ActivePassiveConnectionStrategy`.

dfish3r:
@xweskingx can you try the latest snapshot and confirm it resolves this issue? Thanks.

dfish3r:
No response from reporter.

PR #208, opened 2022-03-21
Add support for building a DNS name from a DN.

dfish3r:
Add DNSDomainFunction which accepts a DN and returns a DNS name.
Update DnsSrvConnectionStrategy to leverage this behavior when it's provided an ldap URL.
See #203.

### DIFF
--- a/core/src/test/java/org/ldaptive/dns/DNSDomainFunctionTest.java
+++ b/core/src/test/java/org/ldaptive/dns/DNSDomainFunctionTest.java
@@ -1,0 +1,74 @@
+/* See LICENSE for licensing and NOTICE for copyright. */
+package org.ldaptive.dns;
+
+import org.ldaptive.dn.Dn;
+import org.ldaptive.dn.NameValue;
+import org.ldaptive.dn.RDn;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * Unit test for {@link DNSDomainFunction}.
+ *
+ * @author  Middleware Services
+ */
+public class DNSDomainFunctionTest
+{
+
+
+  /**
+   * DN test data.
+   *
+   * @return  test data
+   */
+  @DataProvider(name = "DNs")
+  public Object[][] createDNs()
+  {
+    return
+      new Object[][]{
+        new Object[]{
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/ldaptive/dns/DNSDomainFunctionTest.java
```diff
@@ -0,0 +1,74 @@
...
+          "ldaptive.org",
+        },
+        new Object[]{
+          Dn.builder().add("DC=ldap").add("ou=example").add("DC=ldaptive").add("DC=org").build(),
+          "ldaptive.org",
+        },
+        new Object[]{
+          Dn.builder()
+            .add("CN=test")
+            .add(new RDn(new NameValue("DC", "ldap1"), new NameValue("DC", "ldap2")))
```
serac, line 45: I suppose this test case hits the `rdn.size() == 1` condition. Could this be a valid way to say (either "ldap1.ldaptive.org" or "ldap2.ldaptive.org", but the base DN is still "ldaptive.org"? I guess I'm getting at would you expect to see this in the wild or is it more a matter of input validation?

serac:
Looks like it will solve the intended problem nicely. Just had one question mostly for my own understanding of the varied details of DN parsing.

core/src/test/java/org/ldaptive/dns/DNSDomainFunctionTest.java
dfish3r: Not according to the draft:

> An RDN is able to be converted if it (1)
>    consists of a single AttributeTypeAndValue; (2) the attribute type
>    is "DC"; and (3) the attribute value is non-null

So no one should expect that behavior from a multi-value RDN.

serac: Based on the excerpt you cited, should you not throw in that case?

dfish3r: Since non "DC" RDNs are to be expected and ignored, I choose to treat any multi-value RDNs the same. Throwing wouldn't be wrong, but it may result in a more complex implementation. For instance, would you throw for any multi-value RDN or just one containing "DC"?

PR #213, opened 2022-05-16
Set Automatic-Module-Name attributes in MANIFEST.MF

### LINKED ISSUE #212: Basic Java module support
Hi,

It appears that `MANIFEST.MF` doesn't provide the `Automatic-Module-Name` attribute which makes the library difficult to use with the Java module system.

Without it, the name of the module in a `requires` directive would be the name of the JAR file which is subject to change leading to compile and runtime issues.

Would it be possible to include this in the build using something like this:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-jar-plugin</artifactId>
  <configuration>
    <archive>
      <manifestEntries>
        <Automatic-Module-Name>org.ldaptive</Automatic-Module-Name>
      </manifestEntries>
    </archive>
  </configuration>
</plugin>
```

In the long run it would be better to convert the library to a Java module by defining a proper `module-info.java` but the attribute should be enough for now.

Regards

jkuhn1:
Fix #212 

I've updated the `pom.xml` to add the `Automatic-Module-Name` attribute in the `MANIFEST.MF` files

We should now have the following Java automatic modules:

- `org.ldaptive.beans`
- `org.ldaptive.core`
- `org.ldaptive.json`
- `org.ldaptive.templates`

### DIFF
--- a/beans/pom.xml
+++ b/beans/pom.xml
@@ -48,5 +48,21 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  
+  <build>
+  	<plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>org.ldaptive.beans</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
+        </configuration>
+      </plugin>
+  	</plugins>
+  </build>
 
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -46,5 +46,21 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  
+  <build>
+  	<plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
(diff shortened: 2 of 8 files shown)


dfish3r:
Thanks for the patch. I did a bit of reading on the module system and it's not clear to me whether `org.ldaptive.core` or `org.ldaptive` is more appropriate. I guess I'm leaning towards the latter.

jkuhn1:
Well this is similar as to name a Maven artifact since a Java module has a similar semantic. My rule of thumb is to name module after the artifact (group id + artifact id). In this particular case, I looked at the Maven project structure, see `core/`, hence `org.ldaptive.core` but you're right I didn't notice that the artifact ID is actually not `core` but `ldaptive` so `org.ldaptive` might do after all. The only thing to keep in mind is that it has to be unique.

Another important thing to know regarding Java module system, when actually defining a `module-info.java` (automatic modules are exceptions made for compatibility: all packages are exported and this rule is not applied), is the fact that two modules can't export the same package. For example, if you consider modules `org.ldaptive` and `org.ldaptive.beans` only one can export (i.e. make its classes available to other modules) package `org.ldaptive`. It basically forces you to be clear on what a module provides.

dfish3r:
Closed in favor of https://github.com/vt-middleware/ldaptive/commit/4ef6d2eb3b337e593e489ed3f01a4d4fb7478ea7
I riffed on this patch a little bit, but ultimately went with a properties based solution.

PR #222, opened 2022-09-21
Fix the initialization of SyncInfoMessage.

### LINKED ISSUE #221: SyncRepl problems handling with SyncIdSetUuids message
It seems there are some SyncRepl messages resulting into SyncInfoMessage instances providing entryUuids field as null.
Null entryUuids field generate NullPointerExceptions into SyncIdSetUuidsHandler.handle method.
Although I cannot figure out how it is possible to have entryUuids null, I would suggest to provide stronger code into SyncIdSetUuidsHandler.handle. Maybe something like the following

```
        public void handle(final DERParser parser, final DERBuffer encoded) {
            try {
                if (getObject().getEntryUuids()==null) {
                    // ignore and log the skipping
                } else {
                    getObject().addEntryUuids(UuidType.decode(encoded));
                }
            } catch (Exception ignore) {
                // ignore and log the exception
            }
        }
 ```

### PR DESCRIPTION
**Pull request by dfish3r**
IntermediateResponse was designed for subclasses to delegate the BER parsing to it. This can have side effects for subclasses that initialize instance variables. Since the subclass initialization isn't complete until after the superclass construction, any variable setting done by the parser will be over written. Refactor SyncInfoMessage to perform parsing in it's own constructor. Update SyncInfoMessageTest to exercise the public constructor. Fixes #221.

### CODE REVIEW

File: core/src/main/java/org/ldaptive/extended/SyncInfoMessage.java [hunk 1]
```diff
@@ -278,7 +282,7 @@ public void addEntryUuids(final UUID... uuids)
    */
   public void setEntryUuids(final Set<UUID> uuids)
   {
-    entryUuids.addAll(uuids);
+    entryUuids = uuids;
```
serac, line 285: Just noting that copying the input is a little safer to avoid the caller pulling the rug out from under you with a shared reference.

serac (approved):
Looks good. Just a minor suggestion about coding defensively.

core/src/main/java/org/ldaptive/extended/SyncInfoMessage.java
dfish3r: I'm not sure why this method exists. It's not used internally. Comparison with other parsing classes that manage collections indicate that `addFoo` is generally all that is needed.

I've deprecated this method for removal in the next minor release.

PR #227, opened 2022-12-07
Refactor merge to support operation config.

dfish3r:
The underlying operations used by the merge operation can now be configured with handlers. Update MergeOperation to expose properties for these operations. Note that this is a behavioral change to the merge as the initial search is no longer guaranteed to occur on the same connection as the following operation. Add copy methods to each operation for consistency.

### DIFF
--- a/core/src/main/java/org/ldaptive/SearchOperation.java
+++ b/core/src/main/java/org/ldaptive/SearchOperation.java
@@ -827,10 +827,10 @@ public class SearchOperation extends AbstractOperation<SearchRequest, SearchResp
     op.setExceptionHandler(operation.getExceptionHandler());
     op.setThrowCondition(operation.getThrowCondition());
     op.setUnsolicitedNotificationHandlers(operation.getUnsolicitedNotificationHandlers());
+    op.setConnectionFactory(operation.getConnectionFactory());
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/ldaptive/SearchOperation.java
```diff
@@ -827,10 +827,10 @@ public static SearchOperation copy(final SearchOperation operation)
     op.setExceptionHandler(operation.getExceptionHandler());
     op.setThrowCondition(operation.getThrowCondition());
     op.setUnsolicitedNotificationHandlers(operation.getUnsolicitedNotificationHandlers());
+    op.setConnectionFactory(operation.getConnectionFactory());
```
serac, line 830: Was this just for consistency with other copy methods? Does the order matter?

serac (approved):
Looks good.

core/src/main/java/org/ldaptive/SearchOperation.java
dfish3r: Just for consistency, grouping the parent class methods together. Order doesn't matter.

serac: 👍

PR #228, opened 2022-12-14
Add support for batching merge operations.

dfish3r:
Add a flag for controlling whether REPLACE is used versus ADD/DELETE in attribute modifications. Update MergeRequest to include properties for controlling both modification batch size and attribute value batch size. Update MergeOperation to process new MergeRequest properties. Note that the result of the merge operation when batching is now the result of the last operation performed.

### DIFF
--- a/core/src/main/java/org/ldaptive/ext/MergeOperation.java
+++ b/core/src/main/java/org/ldaptive/ext/MergeOperation.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Consumer;
 import org.ldaptive.AddOperation;
 import org.ldaptive.AddRequest;
 import org.ldaptive.AttributeModification;
@@ -12,6 +13,7 @@ import org.ldaptive.Connection;
 import org.ldaptive.ConnectionFactory;
 import org.ldaptive.DeleteOperation;
 import org.ldaptive.DeleteRequest;
+import org.ldaptive.LdapAttribute;
 import org.ldaptive.LdapEntry;
 import org.ldaptive.LdapException;
 import org.ldaptive.ModifyOperation;
@@ -246,7 +248,8 @@ public class MergeOperation
    * @param  source  ldap entry to merge into the LDAP
    * @param  target  ldap entry that exists in the LDAP
    *
-   * @return  response of the modify operation or an empty response if no operation is performed
+   * @return  response of the modify operation or a null response if no operation is performed. If batching is
+   *          enabled in the request, returns the response of the last operation performed
    *
    * @throws  LdapException  if an error occurs executing the modify operation
    */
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/ldaptive/ext/MergeOperation.java
```diff
@@ -402,4 +445,22 @@ protected Result delete(final MergeRequest request, final LdapEntry entry)
...
+
+  /**
+   * Divides the supplied list into sub lists by the supplied divisor and passes each sub list to the consumer.
+   *
+   * @param  <T>  type of list element
+   * @param  list  to divide
+   * @param  divisor  to divide list by
+   * @param  consumer  to process each sub list
+   */
+  private <T> void divideList(final List<T> list, final int divisor, final Consumer<List<T>> consumer)
```
serac, line 458: I wonder whether a `Spliterator` may be more performant here. The usage example at https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Spliterator.html strikes me as awfully similar, and may provide an opportunity for parallelism. Moreover, a dedicated implementation of `Spliterator` may provide a dedicated component to deal with batching concerns. The changes in this MR add a fair bit of complexity to this class soley for batching, and they may be better organized in a separate class.

core/src/main/java/org/ldaptive/ext/MergeOperation.java
```diff
@@ -279,21 +283,60 @@ protected Result modify(
...
-            .modifications(resultModifications.toArray(AttributeModification[]::new))
-            .build());
+        Result result = null;
+        // if batching modifications, execute a modify operation for each batch
+        if (request.getModificationBatchSize() > 0 && request.getModificationBatchSize() < resultModifications.size()) {
+          final List<List<AttributeModification>> batchedResultModifications =
+            new ArrayList<>(resultModifications.size());
+          divideList(resultModifications, request.getModificationBatchSize(), batchedResultModifications::add);
+          for (List<AttributeModification> batch : batchedResultModifications) {
+            final ModifyOperation operation = modifyOperation != null ?
```
serac, line 321: I'm not crazy about cloning the operation to provide batching behavior, but I see that operations invoking other operations is an established pattern so it's probably nothing of concern especially since it's fairly clearly documented in the class javadocs. In general I would think an operation is self-contained, but it appears modify is a special case.

serac (approved):
Looks good with some minor comments/suggestions.

core/src/main/java/org/ldaptive/ext/MergeOperation.java
dfish3r: The clone of the modify is carried over from the last patch, it's not related to this work. The `MergeOperation` carries properties for each add, delete, modify to allow custom configuration to be applied to each operation.

core/src/main/java/org/ldaptive/ext/MergeOperation.java
dfish3r: That's a great suggestion, I'll investigate.

core/src/main/java/org/ldaptive/ext/MergeOperation.java
serac: All good.

core/src/main/java/org/ldaptive/ext/MergeOperation.java
dfish3r: Pushed a commit to reduce complexity in `MergeOperation` by adding a new interface. I didn't get to a `Spliterator` yet, but I think this is moving in the right direction.

serac: Agree, it's a step in the right direction.

PR #240, opened 2023-11-08
Fix WARN logs

leleuj:
On a specific CAS server installation using LDAP via Ldaptive but where the `netty-common` JAR was missing, I have realized that the WARN logs displayed by `NettyConnection` did not output the cause (`java.lang.NoClassDefFoundError`).

This PR fixes the misdefined WARN calls.


dfish3r:
The SLF4J API supports passing an exception as the last parameter to log the stack trace.
https://www.slf4j.org/faq.html#paramException

Were you using an older version of SLF4J?

leleuj:
Good point! The problem comes from the CAS server which uses an old `slf4j-api` version for the 6.6.x stream: https://github.com/apereo/cas/blob/v6.6.13/gradle.properties#L98
Sorry for the noise.

PR #243, opened 2023-12-04
Add timeoutIsFailure and validResultCodes.

dfish3r:
Update SearchConnectionValidator and CompareConnectionValidator to support new configuration options. timeoutIsFailure allows control over whether a connection timeout represents a validation failure. validResultCodes allows control over which result codes represent a validation success.

### DIFF
--- a/core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
+++ b/core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
@@ -32,6 +32,9 @@ public abstract class AbstractConnectionValidator implements ConnectionValidator
   /** Maximum length of time a connection validation should block. */
   private Duration validateTimeout;
 
+  /** Whether the occurrence of a timeout should result in a validation failure. */
+  private boolean timeoutIsFailure = true;
+
 
   @Override
   public Duration getValidatePeriod()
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/ldaptive/AbstractConnectionValidator.java
```diff
@@ -94,7 +119,10 @@ public Supplier<Boolean> applyAsync(final Connection conn)
           // waits indefinitely for the validation response
           latch.await();
         } else {
-          latch.await(getValidateTimeout().toMillis(), TimeUnit.MILLISECONDS);
```
serac, line 97: Since the code ignores the return value, the previous behavior is unclear. Can you please explain what would happen in that case with the original code?

dfish3r: You're waiting on the function passed to #applyAsync to give you a boolean. It's up to that implementation to determine when that occurs, but generally it's after an LDAP operation has occurred. So the original version simply used the default value of the `AtomicBoolean` (which is false) when it is done waiting on the `latch`.

serac: I see, thanks.

serac (approved):
Looks like a good enhancement for increased flexibility in handling connection validation failures including timeouts.

PR #244, opened 2023-12-10
Add validator to SingleConnectionFactory.

dfish3r:
The SyncReplRunner attempted to maintain a valid connection using exception handling. This proved to be difficult to configure and complex to implement. Using an external thread to ensure the connection is valid has proven to be a simplier and more reliable solution.

Add validator, onOpen and onClose properties to SingleConnectionFactory. Add onSuccess and onFailure properties to AbstractConnectionValidator. Update SingleConnectionFactory to execute the validator when it is configured. Add a ReinitializeConnectionConsumer that SyncReplRunner can leverage to maintain its connection. Update SyncReplRunner to use a ConnectionValidator instead of exception handling.

### DIFF
--- a/core/src/main/java/org/ldaptive/SingleConnectionFactory.java
+++ b/core/src/main/java/org/ldaptive/SingleConnectionFactory.java
@@ -7,6 +7,10 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.Function;
 import org.ldaptive.transport.Transport;
 import org.ldaptive.transport.TransportFactory;
 
@@ -18,11 +22,8 @@ import org.ldaptive.transport.TransportFactory;
 public class SingleConnectionFactory extends DefaultConnectionFactory
 {
 
-  /** The connection used by this factory. */
-  private Connection connection;
-
   /** The proxy used by this factory. */
-  private Connection proxy;
+  private ConnectionProxy proxy;
 
   /** Whether {@link #initialize()} has been successfully invoked. */
   private boolean initialized;
@@ -33,8 +34,17 @@ public class SingleConnectionFactory extends DefaultConnectionFactory
   /** Whether {@link #initialize()} should occur on a separate thread. */
   private boolean nonBlockingInitialize;
 
-  /** Executor for scheduling {@link #initialize()}. */
-  private ExecutorService initializeExecutor;
+  /** To run when a connection is opened. */
+  private Function<Connection, Boolean> onOpen;
+
+  /** To run when a connection is closed. */
+  private Function<Connection, Boolean> onClose;
+
+  /** For validating the connection. */
+  private ConnectionValidator validator;
+
+  /** Executor for scheduling factory tasks. */
+  private ExecutorService factoryExecutor;
 
 
   /** Default constructor. */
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/ldaptive/SingleConnectionFactory.java
```diff
@@ -366,9 +518,30 @@ protected Builder(final Transport transport)
...
 
-    public Builder config(final ConnectionConfig cc)
+    public Builder config(final ConnectionConfig config)
+    {
+      object.setConnectionConfig(config);
+      return this;
+    }
+
+
+    public Builder onOpen(final Function<Connection, Boolean> function)
```
serac, line 528: These methods seem reasonable in general, but I see that the `SyncReplRunner` use case drove their creation. Can you imagine another use case? Just trying to press on the API change a bit to justify it.

core/src/main/java/org/ldaptive/control/util/SyncReplRunner.java
```diff
@@ -94,44 +65,136 @@ public void accept(final Exception e)
...
+
+  /**
+   * Creates a new single connection factory. Uses a {@link SearchConnectionValidator} for connection validation. See
+   * {@link #createTransport()}
+   *
+   * @param  config  sync repl connection configuration
+   *
+   * @return  single connection factory for use with a sync repl runner
+   */
+  public static SingleConnectionFactory createConnectionFactory(final ConnectionConfig config)
```
serac, line 120: While I can imagine these are helpful, their presence in this class strikes me as odd. Consider a separate class with convenience constructors; compare with the `Executors` JDK class.

core/src/main/java/org/ldaptive/control/util/SyncReplRunner.java
```diff
@@ -94,44 +65,136 @@ public void accept(final Exception e)
...
+   *   <li>{@link ConnectionConfig#setAutoReplay(boolean)} to false</li>
+   *   <li>{@link SingleConnectionFactory#setFailFastInitialize(boolean)} to false</li>
+   *   <li>{@link SingleConnectionFactory#setNonBlockingInitialize(boolean)} to false</li>
+   *   <li>{@link AbstractConnectionValidator#setOnFailure(Consumer)} to
+   *   {@link SingleConnectionFactory.ReinitializeConnectionConsumer}</li>
+   * </ul>
+   *
+   * @param  factory  to configure
+   */
+  public static void configureConnectionFactory(final SingleConnectionFactory factory)
```
serac, line 183: This is the only static method that seems naturally at home in this class since the configuration behavior is tailed specifically to `SyncReplRunner`.

serac (approved):
Looks like it solves the problem stated in the summary, though I must admit that I'm not clear on exactly _why_ managing this component's connection via exceptions was so difficult. That said I trust you had sufficient evidence.

core/src/main/java/org/ldaptive/SingleConnectionFactory.java
dfish3r: I think any use case that involves a persistent search. Content sync isn't the only RFC out there, just the only one we've attempted to use. For instance, AD has the dir sync control.

core/src/main/java/org/ldaptive/control/util/SyncReplRunner.java
dfish3r: Would a new class called `ConnectionFactories` make more sense? With methods like:
- `ConnectionFactories#newSyncReplRunnerFactory(ConnectionConfig)`
- `ConnectionFactories#newSyncReplRunnerFactory(ConnectionConfig, ConnectionValidator)`
- `ConnectionFactories#newSyncReplRunnerFactory(Transport, ConnectionConfig, ConnectionValidator)`

serac: Yes, that's in line with my thinking.

dfish3r: Thinking on this some more, I'm not inclined to add a separate helper class _yet_. You already noted that we only have a single use case. I'd prefer for one or two more use cases to arrive before moving those methods out of the `SyncReplRunner`. They arguably make more sense where they are today.

PR #250, opened 2024-02-13
Change the behavior of response timeout.

### LINKED ISSUE #248: Response timeout is too strict for search operations
The implementation of operation timeouts waits until an entire response is received. For search operations the timeout should support references and entries.

dfish3r:
Change from a countDownLatch to a semaphore to support alternate use cases. Use a Predicate to determine whether to throw a timeout exception. The previous implementation only accepted a Result message. Add some sanity checks to message processing.
Fixes #248.

### DIFF
--- a/core/src/main/java/org/ldaptive/transport/DefaultCompareOperationHandle.java
+++ b/core/src/main/java/org/ldaptive/transport/DefaultCompareOperationHandle.java
@@ -146,6 +146,12 @@ public class DefaultCompareOperationHandle
    */
   public void compare(final CompareResponse response)
   {
+    if (getMessageID() != response.getMessageID()) {
+      final IllegalArgumentException e = new IllegalArgumentException(
+        "Invalid compare response " + response + " for handle " + this);
+      exception(new LdapException(e));
+      throw e;
+    }
     if (onCompare != null) {
       for (CompareValueHandler func : onCompare) {
         try {
--- a/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
+++ b/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
@@ -665,21 +665,17 @@ public class DefaultOperationHandle<Q extends Request, S extends Result> impleme
       }
     }
     exception = e;
-    responseSemaphore.release();
+    consumedMessage();
     complete();
   }
 
 
   /**
    * Indicates that a protocol message was consumed by a supplied consumer.
-   *
-   * @deprecated  Use {@link #consumedMessage(Message)}
    */
-  @Deprecated
   protected void consumedMessage()
   {
-    consumedMessage = true;
-    responseSemaphore.release();
+    consumedMessage(true);
   }
 
 
@@ -690,8 +686,19 @@ public class DefaultOperationHandle<Q extends Request, S extends Result> impleme
    */
   protected void consumedMessage(final Message message)
   {
+    consumedMessage(getResponseTimeoutCondition().test(message));
+  }
+
+
+  /**
+   * Indicates that a protocol message was consumed by a supplied consumer.
+   *
+   * @param  signalResponseSemaphore  whether to signal the response semaphore
+   */
+  private void consumedMessage(final boolean signalResponseSemaphore)
+  {
     consumedMessage = true;
-    if (getResponseTimeoutCondition().test(message)) {
+    if (signalResponseSemaphore) {
       responseSemaphore.release();
     }
   }
--- a/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
+++ b/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
@@ -3,20 +3,37 @@ package org.ldaptive.transport;
 
 import java.time.Duration;
 import java.time.Instant;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 import org.ldaptive.AbandonRequest;
+import org.ldaptive.AddRequest;
+import org.ldaptive.AddResponse;
 import org.ldaptive.BindRequest;
+import org.ldaptive.BindResponse;
+import org.ldaptive.CompareRequest;
+import org.ldaptive.CompareResponse;
+import org.ldaptive.DeleteRequest;
+import org.ldaptive.DeleteResponse;
 import org.ldaptive.LdapException;
+import org.ldaptive.Message;
+import org.ldaptive.ModifyDnRequest;
+import org.ldaptive.ModifyDnResponse;
+import org.ldaptive.ModifyRequest;
+import org.ldaptive.ModifyResponse;
 import org.ldaptive.OperationHandle;
 import org.ldaptive.Request;
 import org.ldaptive.Result;
 import org.ldaptive.ResultCode;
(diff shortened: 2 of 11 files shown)


core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
```diff
@@ -111,6 +147,24 @@ public class DefaultOperationHandle<Q extends Request, S extends Result> impleme
...
+   * @param  req  request to expect a response for
+   * @param  conn  the request will be executed on
+   * @param  timeout  duration to wait for a response
+   * @param  timeoutCondition  whether a message should reset the timeout
+   */
+  DefaultOperationHandle(
+    final Q req,
+    final TransportConnection conn,
+    final Duration timeout,
+    final Predicate<Message> timeoutCondition)
```
serac, line 167: I can't imagine any value to exposing this condition to your API, and yet I see costs. Moving the signaling out of `complete` into two distinct methods, `completeMessage` and `exception`, is harder to follow and _feels_ riskier. You want to guarantee that the semaphore signals on both errors and responses, and the more you disperse that signaling the harder it is to reason through the concurrency as well as ensure that it actually happens. I'm not sure you've gained enough for that increase in complexity/risk.

core/src/main/java/org/ldaptive/transport/DefaultCompareOperationHandle.java
```diff
@@ -146,6 +146,12 @@ public DefaultCompareOperationHandle onCompare(final CompareValueHandler... func
    */
   public void compare(final CompareResponse response)
   {
+    if (getMessageID() != response.getMessageID()) {
```
serac, line 149: I don't understand the need for this check. Is it needed for unsolicited responses?

core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
```diff
@@ -567,40 +676,55 @@ public void exception(final LdapException e)
...
+    }
   }
 
 
   /**
    * Releases the latch and sets the response as received. Invokes {@link #onComplete}. Handle is considered done when
    * this is invoked.
    */
-  private void complete()
+  private synchronized void complete()
```
serac, line 715: I don't understand the need to make this method synchronized. The same instance variables are at play, so it seems you don't need method synchronization unless you feel you missed something previously.

core/src/main/java/org/ldaptive/transport/DefaultSearchOperationHandle.java
```diff
@@ -30,6 +33,22 @@ public class DefaultSearchOperationHandle
   extends DefaultOperationHandle<SearchRequest, SearchResponse> implements SearchOperationHandle
 {
 
+  /** Predicate that requires any message except unsolicited. */
+  protected static final Predicate<Message> SEARCH_RESPONSE_TIMEOUT_CONDITION =
```
serac, line 37: It's not clear why `DEFAULT_RESPONSE_TIMEOUT_CONDITION` is insufficient for this case.

serac (approved):
This looks like a solid improvement to implement a running timeout that sets the response timeout based on the _last_ result received rather than some absolute timeout, which seems much more consistent with expectations. I had a relatively minor API concern, and everything else consisted of questions for my own understanding.

core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
serac: If you truly need two different predicates, a default and a distinct one for search, then you'd be better served in my opinion by using a protected abstract template method that returns the correct value based on context. In that way it would not be exposed to callers as an API input.

dfish3r: I tend to agree. I was leaving myself some rope in case we wanted to support the old behavior for searches. I'll push a fix that uses a `protected` method.

core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
dfish3r: I missed something previously. There is a potential race condition where a result and an exception could occur in a timespan that would allow two separate threads to enter this method. This could potentially result in `onComplete` and/or `connection#complete` being invoked twice.

core/src/main/java/org/ldaptive/transport/DefaultSearchOperationHandle.java
dfish3r: The purpose of this patch is to treat search results differently from other operations. We want the `responseTimeout` to restart when other messages are received, not just the `Result` type. Would this be clearer if I narrowed the type checking?

core/src/main/java/org/ldaptive/transport/DefaultCompareOperationHandle.java
serac: Understood, thanks for the explanation.

core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
serac: Understood, good catch.

core/src/main/java/org/ldaptive/transport/DefaultSearchOperationHandle.java
serac: The question arose from ignorance of how LDAP responses are mapped to ldaptive API types. No changes needed, thanks for clarifying.

serac:
The changes to hide the response timeout condition a bit look good. I took the liberty of playing around with consolidating the signaling logic in a single method and came up with something, but in fairness it's only an academic improvement:

      diff --git a/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java b/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
      index 19078026e..af78a08d2 100644
      --- a/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
      +++ b/core/src/main/java/org/ldaptive/transport/DefaultOperationHandle.java
      @@ -665,21 +665,17 @@ public class DefaultOperationHandle<Q extends Request, S extends Result> impleme
            }
         }
         exception = e;
      -    responseSemaphore.release();
      +    consumedMessage();
         complete();
         }
      
      
         /**
         * Indicates that a protocol message was consumed by a supplied consumer.
      -   *
      -   * @deprecated  Use {@link #consumedMessage(Message)}
         */
      -  @Deprecated
         protected void consumedMessage()
         {
      -    consumedMessage = true;
      -    responseSemaphore.release();
      +    consumedMessage(true);
         }
      
      
      @@ -689,9 +685,19 @@ public class DefaultOperationHandle<Q extends Request, S extends Result> impleme
         * @param  message  that was consumed
         */
         protected void consumedMessage(final Message message)
      +  {
      +    consumedMessage(getResponseTimeoutCondition().test(message));
      +  }
      +
      +  /**
      +   * Marks a message consumed and optionally signals the response semaphore if the predicate evaluates to true.
      +   *
      +   * @param signalResponseSemaphore True to signal the response semaphore, false otherwise.
      +   */
      +  private void consumedMessage(final boolean signalResponseSemaphore)
         {
         consumedMessage = true;
      -    if (getResponseTimeoutCondition().test(message)) {
      +    if (signalResponseSemaphore) {
            responseSemaphore.release();
         }
         }

dfish3r:
@serac I pushed your recommendation. I think this is ready for merge.

PR #253, opened 2024-04-04
Fix multi-byte assertion values.

### LINKED ISSUE #252: Problems with encoding using the search filter
While all the information on LDAP server consist of latin charachters all seems to be working just fine.

But if there exist some different charachters I couldn't get search request working.

Let's say we're looking for user dn like this:
```
        val userSearchRequest = SearchRequest.builder()
            .dn("dc=company,dc=com")
            .filter("(uid=username)")
            .sizeLimit(1)
            .scope(SearchScope.SUBTREE)
            .build()

        val userDn = SearchOperation(connectionFactory).execute(userSearchRequest)
            .entry
            .dn
```

And we got perfectly fine dn in UTF-8, but with non latin characters present:
`uid=username,ou=Лаборатория,ou=Университет,dc=company,dc=com`

Next we're doing search which groups user is a member of:
```
        val groupsUserMemberInRequest = SearchRequest.builder()
            .dn("dc=company,dc=com")
            .filter("(member=$userDn)") //here i can put "uid=username,ou=Лаборатория,ou=Университет,dc=company,dc=com" in UTF-8 as a string instead of $userDn, it won't make any difference
            .scope(SearchScope.SUBTREE)
            .build()

        val groups = SearchOperation(connectionFactory).execute(groupsUserMemberInRequest)
            .entries
```

And I got nothing (same request works just fine in Apache Directory Studio).
I assume something wrong with encoding in the filter, because when I debug this code, I see the following:

`org.ldaptive.SearchRequest@-463502941::controls=null, dn=DC=company,DC=com, scope=SUBTREE, aliases=NEVER, sizeLimit=0, timeLimit=PT0S, typesOnly=false, filter=org.ldaptive.filter.EqualityFilter@-1616026601::filterType=EQUALITY, attributeDesc=member, assertionValue=uid=username,ou=01>@0B>@8O,ou=#=825@A8B5B,dc=company,dc=com, returnAttributes=[*], binaryAttributes=null`

Assertion value doesn't seem to be right:
`assertionValue=uid=username,ou=01>@0B>@8O,ou=#=825@A8B5B,dc=company,dc=com`

Can you please advice what I can do about this issue?

dfish3r:
FilterUtils was only writing the first byte of multi-byte characters. Fixes #252.

### DIFF
--- a/core/src/main/java/org/ldaptive/filter/FilterUtils.java
+++ b/core/src/main/java/org/ldaptive/filter/FilterUtils.java
@@ -2,6 +2,7 @@
 package org.ldaptive.filter;
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import org.ldaptive.LdapUtils;
 import org.ldaptive.ResultCode;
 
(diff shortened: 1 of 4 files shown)


dhawes:
Looks good. This is a nice table for anyone trying to visualize the bitwise and:

https://en.wikipedia.org/wiki/UTF-8#Encoding

PR #254, opened 2024-04-05
Add max depth to filter parsing.

dfish3r:
Prevent stack overflows by limiting the depth of filter parsing.


serac:
Admittedly defensive, but given the ease of implementation for a little extra protection seems worthwhile.

PR #255, opened 2024-04-17
Add spotbugs maven plugin.

dfish3r:


serac:
The issues the plugin found around synchronization may merit further study as time allows. I'm confident there is sufficient testing and usage in practice to demonstrate proper behavior, but perhaps there is room for improvement in clarity.

PR #256, opened 2024-04-19
Separate LDAP URL parsing from state management.

dfish3r:
Bump minor version for API breaking changes in LdapURL. Provide a separate package for parsing and formatting LDAP URLs. Add default Url parsing implementation.
Change LdapURL to use composition for parsed URL properties. Fix percent encode/decode for multi-byte characters. Add benchmark tests for default and regex URL parsing.


core/src/main/java/org/ldaptive/url/DefaultUrlParser.java
```diff
@@ -0,0 +1,232 @@
...
+    final int port;
+    if (colonPos < 0) {
+      hostname = hostAndPort;
+      port = -1;
+    } else {
+      hostname = hostAndPort.substring(0, colonPos);
+      port = Integer.parseInt(hostAndPort.substring(colonPos + 1));
+      Url.validatePort(port, false);
+    }
+    return new Object[] {hostname.isEmpty() ? null : hostname, port};
```
serac, line 155: Seems like you should throw here rather than returning null. I can't imagine a URL without a hostname is a valid condition.

core/src/main/java/org/ldaptive/url/Url.java
```diff
@@ -0,0 +1,443 @@
...
+  private final String hostname;
+
+  /** Port of the ldap url. */
+  private final int port;
+
+  /** Base DN of the ldap url. */
+  private final String baseDn;
+
+  /** Parsed base DN of the ldap url. */
+  private final Dn parsedBaseDn;
```
serac, line 63: You are representing the DN twice in the underlying data model. While I can appreciate that storing both the parse and unparsed representations may be helpful, you should not expose this detail externally. Pick one for your constructors and derive the other if necessary.

core/src/main/java/org/ldaptive/url/Url.java
```diff
@@ -0,0 +1,443 @@
...
+  private final String[] attributes;
+
+  /** Search scope of the ldap url. */
+  private final SearchScope scope;
+
+  /** Search filter of the ldap url. */
+  private final String filter;
+
+  /** Parsed filter of the ldap url. */
+  private final Filter parsedFilter;
```
serac, line 75: Same concern as DN. It's ok to represent both forms internally as an optimization/convenience/implementation detail, but both forms should not be exposed to construct the object.

core/src/main/java/org/ldaptive/url/Url.java
```diff
@@ -0,0 +1,443 @@
...
+    return formatter.format(this);
+  }
+
+
+  /**
+   * Determines whether the supplied scheme is valid. Must be one of 'ldap', 'ldaps' or 'ldapi'.
+   *
+   * @param  scheme  to validate
+   */
+  static void validateScheme(final String scheme)
```
serac, line 362: I have a slight preference that these validations live in a utility class. I believe `LdapUtils` is a suitable existing location.

core/src/main/java/org/ldaptive/url/RegexUrlParser.java
```diff
@@ -0,0 +1,118 @@
...
+import org.ldaptive.filter.Filter;
+import org.ldaptive.filter.FilterFunction;
+import org.ldaptive.filter.RegexFilterFunction;
+
+/**
+ * Parses an LDAP URL string using a regular expression.
+ *
+ * @author  Middleware Services
+ */
+public final class RegexUrlParser implements UrlParser
```
serac, line 21: Strongly recommend you provide exactly one implementation of `UrlParser` with the library. It's my understanding that `DefaultUrlParser` is more performant; perhaps that should be the implementation you ship.

serac (commented):
This looks good. At first glance I wondered why you didn't simply use the JDK `URI` class, but I can see upon review that your custom `Url` class is fit for the needs of the library and makes sense. I had a couple concerns about duplication of data leaking out into the API, but other than that looks good.

core/src/main/java/org/ldaptive/url/DefaultUrlParser.java
dfish3r: I made it valid for the `UrlParser` implementation, but invalid for `LdapURL`. Technically host is not required: https://datatracker.ietf.org/doc/html/rfc4516#section-3

core/src/main/java/org/ldaptive/url/Url.java
dfish3r: How the DN is parsed is actually an implementation detail for the `UrlParser`, which is why it's provided and not derived. I really wanted a single parameter, like a supplier that produces a tuple. But that produced more code than it was worth.

core/src/main/java/org/ldaptive/url/RegexUrlParser.java
dfish3r: It becomes the default implementation with this patch. I'd prefer not to throw away the regex parser just yet, probably in a future release.

serac: Suggest you leverage `@Deprecated` if you thinking of future removal.

core/src/main/java/org/ldaptive/url/DefaultUrlParser.java
serac: I just realized that you already have a class, `LdapURL` to describe LDAP URLs. You ought to have exactly one class for this concern in the library and use it consistently. I can see that `LdapURL` perhaps needs to continue to exist as an API bridge, but you should move to consolidate on a single data model. Then I will buy the argument that it can be null, as it will be nullable everywhere.

core/src/main/java/org/ldaptive/url/Url.java
serac: I feel pretty strongly that this is a sign of a need for refinement. Taking what you said as fact, then I would augment the `UrlParser` API to address DN parsing in addition to LDAP URL parsing. Then you could use composition in your method calls to avoid duplication. I'm squinting to see the solution, but my instinct says that's one way to keep concerns as you have them and avoid this duplication.

core/src/main/java/org/ldaptive/url/DefaultUrlParser.java
dfish3r: One goal of this patch is to separate the concerns of LDAP connections from URL parsing. The `url` package was created to address the latter. The `LdapURL` class was refactored to address the former. Since `LdapURL` contains state data specific to a TCP connection and is produced by `ConnectionStrategy` it made more sense for it to have an instance of an parsed URL to use for its purposes.

Meanwhile, any use cases that require URL parsing can use the `url` package.

core/src/main/java/org/ldaptive/url/Url.java
dfish3r: It sounds like you'd prefer the `Url` interface to change from:
```
Url parse(String url);
````
to
```
Url parse(String url, DnParser dnParser, FilterFunction filterFunction);
```

dfish3r: I can also remove the parsed fields and leave parsing as an exercise for the `UrlFormatter` implementations. That's probably the cleanest solution.

core/src/main/java/org/ldaptive/url/DefaultUrlParser.java
serac: I encourage you to consider consolidation at some point in the future.###

core/src/main/java/org/ldaptive/url/Url.java
serac: I think your instinct is right, but I want to avoid over-engineering here. You don't have a use case for the parsed DN given the `UrlFormatter` you provide with the library, `MinimalUrlFormatter`. If you feel strongly that working on `Dn` object may be useful at some point, you'd be better off generifying the interface to accept an arbitrary type for the DN, and then providing `MinimalUrlFormatter<String>` with a string representation of the DN. The you've given rope to provide a `CustomUrlFormatter<Dn>` in the future. All that said, is it worth it?

sysmat: it is better to have strong types, String as dns is not valid, but as Dn then you know is valid

dfish3r: I ended up internalizing the parsing of both the `parsedDn` and the `parsedFilter`. So `Url` can expose both the original version and the parsed version. It's not the cleanest implementation, but I think it's the most flexible. @serac have a look at the latest.

core/src/main/java/org/ldaptive/url/Url.java
dfish3r: Resolving this as part of the `parsedDn` conversation.

core/src/main/java/org/ldaptive/url/RegexUrlParser.java
dfish3r: Made a note for future deprecation.

core/src/main/java/org/ldaptive/url/Url.java
serac: This addresses my biggest concern, so we can move forward.

core/src/main/java/org/ldaptive/url/Url.java
dfish3r: These are very specific to the `Url` implementations, I'd prefer not to expose them generally until a general use case arises.

PR #257, opened 2024-05-08
Prefer byte array over ByteBuffer for properties.

dfish3r:
Add AttributeValue to reduce complexity of LdapAttribute. Copy input and output byte arrays to guard against modification. Don't allow setting of null attribute type or value. Improve LdapAttribute test coverage.

### DIFF
--- a/core/src/main/java/org/ldaptive/LdapAttribute.java
+++ b/core/src/main/java/org/ldaptive/LdapAttribute.java
@@ -1,7 +1,7 @@
 /* See LICENSE for licensing and NOTICE for copyright. */
 package org.ldaptive;
 
-import java.nio.ByteBuffer;
+import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


serac (approved):
I think using immutable types for public APIs is generally beneficial, and the switch to `byte[]` here is consistent with that practice. Additionally, I can see some performance/efficiency improvements with these changes.

core/src/main/java/org/ldaptive/LdapAttribute.java
```diff
@@ -591,68 +551,220 @@ public String toString()
...
+    private final byte[] value;
+
+
+    /**
+     * Creates a new attribute value.
+     *
+     * @param bytes byte array value
+     * @param copy  whether to copy the supplied bytes
+     */
+    AttributeValue(final byte[] bytes, final boolean copy)
```
serac, line 614: I didn't see any usage of this constructor with the copy flag set to true. Did I miss it?

dfish3r: Yes, `addValues` and `addBinaryValues` both use it.

PR #258, opened 2024-05-15
Propagate immutable flag into all config.

dfish3r:
Expand the use immutable into all configuration objects. Add Immutable and AbstractImmutable classes.
Update immutable objects to check for immutable on mutation. For array type configuration objects, implement copy protection on both getters and setters. Return unmodifiable collections from getters where appropriate. Review use of the final keyword for various extension points.


serac:
I am generally in favor of this approach and think the use of an interface to distinguish "immutable after configuration" is a very useful and clarifying concept. My only concern is the use of the name `Immutable` for this type, and I took the liberty to propose an alternative, `Freezable`: https://github.com/vt-middleware/ldaptive/pull/259.

serac (approved):
I'm generally in favor of this approach and really like the use of a type to describe the concept of "immutable after configuration."

PR #259, opened 2024-05-17
Rename Immutable -> Freezable.

serac:
The concept of "immutable after a method invocation" is substantially different from the typical sense of an immutable type and as such merits a distinguishing name: Freezable.


core/src/main/java/org/ldaptive/AbstractFreezable.java
```diff
@@ -6,29 +6,29 @@
  *
  * @author  Middleware Services
  */
-public abstract class AbstractImmutable implements Immutable
+public abstract class AbstractFreezable implements Freezable
 {
 
   /** Whether this object has been marked immutable. */
   private volatile boolean immutable;
```
dfish3r, line 13: Rename to `frozen` ?

core/src/main/java/org/ldaptive/AbstractFreezable.java
```diff
@@ -6,29 +6,29 @@
...
-  public final boolean isImmutable()
+  public final boolean isFrozen()
   {
     return immutable;
   }
 
 
   @Override
-  public final void checkImmutable()
+  public final void assertMutable()
```
dfish3r, line 31: Why not `assertFrozen()` ?

PR #260, opened 2024-05-21
Add freeze interface to entry and result handlers.

dfish3r:
Update mutable methods to be final and use assertMutable. Refactor heirarchy of handlers that don't need MessageFunctional. Add newInstance to MessageFunctional so that handlers can be mutated by the operation handle. Add convenience constructors and builders to some handlers. Add additional unit tests.

### DIFF
--- a/core/src/main/java/org/ldaptive/handler/CaseChangeEntryHandler.java
+++ b/core/src/main/java/org/ldaptive/handler/CaseChangeEntryHandler.java
@@ -11,7 +11,7 @@ import org.ldaptive.LdapUtils;
  *
  * @author  Middleware Services
  */
-public class CaseChangeEntryHandler extends AbstractEntryHandler<LdapEntry> implements LdapEntryHandler
+public class CaseChangeEntryHandler extends AbstractEntryHandler implements LdapEntryHandler
 {
 
   /** hash code seed. */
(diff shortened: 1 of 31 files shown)


core/src/main/java/org/ldaptive/handler/CaseChangeEntryHandler.java
```diff
@@ -245,4 +267,109 @@ public String toString()
...
+     */
+    private Builder() {}
+
+
+    /**
+     * Makes this instance immutable.
+     *
+     * @return  this builder
+     */
+    public Builder freeze()
```
serac, line 302: It strikes me that the object should be implicitly frozen upon calling `build()`, which would

serac, line 302: It strikes me that the object should be implicitly frozen upon calling `build()`, which would remove the need for this public method.

core/src/main/java/org/ldaptive/handler/DnAttributeEntryHandler.java
```diff
@@ -118,4 +137,81 @@ public String toString()
...
+     */
+    private Builder() {}
+
+
+    /**
+     * Makes this instance immutable.
+     *
+     * @return  this builder
+     */
+    public Builder freeze()
```
serac, line 172: Same concern as above.

core/src/main/java/org/ldaptive/handler/MergeAttributeEntryHandler.java
```diff
@@ -133,4 +152,81 @@ public String toString()
...
+     */
+    private Builder() {}
+
+
+    /**
+     * Makes this instance immutable.
+     *
+     * @return  this builder
+     */
+    public Builder freeze()
```
serac, line 187: Same concern as above.

core/src/main/java/org/ldaptive/transport/MessageFunctional.java
```diff
@@ -1,21 +1,22 @@
...
+ * Base class for processing a message that is initialized with properties from the handle.
  *
  * @param  <Q> type of request
  * @param  <S> type of result
  *
  * @author  Middleware Services
  */
 // CheckStyle:AbstractClassName OFF
-public abstract class MessageFunctional<Q extends Request, S extends Result>
+public abstract class MessageFunctional<Q extends Request, S extends Result> extends AbstractFreezable
```
serac, line 19: You have a bit of a naming issue here with an abstract base class that doesn't begin with the typical `Abstract` name. May I suggest `AbstractMessageBasedResultHandler`.

serac (approved):
Looks good. Had a couple of minor suggestions.

core/src/main/java/org/ldaptive/transport/MessageFunctional.java
dfish3r: I'll take that under advisement for a future patch.

PR #262, opened 2024-05-24
Update response objects to support freeable.

dfish3r:
Add concept of immutableOnConstruct so that response objects have base properties that cannot be changed by handlers and other properties that can be changed. Add static copy methods so that new instances of objects can be created that are mutable.

### DIFF
--- a/core/src/main/java/org/ldaptive/AbstractMessage.java
+++ b/core/src/main/java/org/ldaptive/AbstractMessage.java
@@ -33,18 +33,49 @@ import org.ldaptive.control.ResponseControl;
  *
  * @author  Middleware Services
  */
-public abstract class AbstractMessage implements Message
+public abstract class AbstractMessage extends AbstractFreezable implements Message
 {
+
+  /** LDAP controls. */
+  private final List<ResponseControl> controls = new ArrayList<>();
 
   /** Protocol message ID. */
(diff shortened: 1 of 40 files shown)


serac (changes requested):
While I am supportive of the _concept_ of immutable responses, I feel that you have gone too far in trying to adapt the `Freezable` type to this use case. After review and reflection, I feel that `Freezable` is best applied to configuration classes where it is natural to configure objects via setters and then invoke some method to indicate "done" that prevents subsequent modification. For core data types, on the other hand, I feel that naturally immutable types are simpler, clearer, and perhaps even safer, where "naturally immutable" means that the objects provide no setters and all fields that describe state are required to be provided by in the class constructor (and are consequently final). A builder pattern is needed to support construction in order to build up the objects as data is obtained and then supplying all the data in one shot to the constructor. I have created a branch that implements this approach for just _some_ of the classes that implement `AbstractResult`:

https://github.com/vt-middleware/ldaptive/tree/immutable-response-with-builders

The use of lombok to generate builders appears to be a promising approach to avoid a dramatic increase of boilerplate code that would otherwise be needed with this strategy.

dfish3r:
Closed in favor on https://github.com/vt-middleware/ldaptive/pull/263.

PR #263, opened 2024-06-04
Update search objects to support freezable.

dfish3r:
Update objects used in the search response to implement Freezable. Update DefaultSearchOperationHandle to freeze the SearchResponse on return. Add static copy methods so that new instances of objects can be created that are mutable. Change response setter methods visibility to enforce immutability on response objects.

### DIFF
--- a/core/src/main/java/org/ldaptive/LdapEntry.java
+++ b/core/src/main/java/org/ldaptive/LdapEntry.java
@@ -3,6 +3,7 @@ package org.ldaptive;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -35,7 +36,7 @@ import org.ldaptive.dn.Dn;
  *
  * @author  Middleware Services
  */
-public class LdapEntry extends AbstractMessage
+public class LdapEntry extends AbstractMessage implements Freezable
 {
 
   /** BER protocol number. */
@@ -50,6 +51,9 @@ public class LdapEntry extends AbstractMessage
   /** DER path to attributes. */
   private static final DERPath ATTRIBUTES_PATH = new DERPath("/SEQ/APP(4)/SEQ/SEQ");
 
+  /** LDAP attributes on the entry. */
+  private final Map<String, LdapAttribute> attributes = new LinkedHashMap<>();
+
   /** LDAP DN of the entry. */
   private String ldapDn;
 
@@ -59,8 +63,8 @@ public class LdapEntry extends AbstractMessage
   /** Normalized LDAP DN. */
   private String normalizedDn;
 
-  /** LDAP attributes on the entry. */
-  private Map<String, LdapAttribute> attributes = new LinkedHashMap<>();
+  /** Whether this object has been marked immutable. */
+  private volatile boolean immutable;
 
 
   /**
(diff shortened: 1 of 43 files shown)


core/src/main/java/org/ldaptive/LdapEntry.java
```diff
@@ -59,8 +63,8 @@ public class LdapEntry extends AbstractMessage
   /** Normalized LDAP DN. */
   private String normalizedDn;
 
-  /** LDAP attributes on the entry. */
-  private Map<String, LdapAttribute> attributes = new LinkedHashMap<>();
+  /** Whether this object has been marked immutable. */
+  private volatile boolean immutable;
```
serac, line 67: Why volatile?

core/src/main/java/org/ldaptive/auth/AuthenticationResponse.java
```diff
@@ -50,6 +50,7 @@ public AuthenticationResponse(
     authenticationHandlerResponse = response;
     resolvedDn = dn;
     ldapEntry = entry;
+    ldapEntry.freeze();
```
serac, line 53: I appreciate this change and think it's a meaningful move toward immutability for authentication-related data. Ideally all data in the authentication response, including the account state, would be immutable.

core/src/main/java/org/ldaptive/handler/RecursiveResultHandler.java
```diff
@@ -196,19 +203,18 @@ private void readSearchAttribute(final LdapEntry entry, final List<String> searc
...
-      try {
-        final SearchResponse result = performSearch(dn, retAttrs);
-        if (result.isSuccess() && result.entrySize() == 1) {
-          newEntry = result.getEntry();
-        }
-      } catch (LdapException e) {
-        logger.warn("Error retrieving attribute(s): {} for dn {}", Arrays.toString(retAttrs), dn, e);
+      final SearchResponse result = performSearch(dn, retAttrs);
+      if (result.isSuccess() && result.entrySize() == 1) {
+        newEntry = LdapEntry.copy(result.getEntry());
```
serac, line 217: I don't see the value of making a copy of the result here.

serac (approved):
This is a good step toward making response types immutable in general with the special case of search results, which are intended to support mutation up to the point they're frozen.

core/src/main/java/org/ldaptive/LdapEntry.java
dfish3r: Perhaps overzealous thinking on my part to make data consistency safe in multi-threaded environments. You could convince me this isn't needed since other properties aren't thread safe. But maybe this flag should be given special consideration.

core/src/main/java/org/ldaptive/auth/AuthenticationResponse.java
dfish3r: I have an idea to deal with that in a subsequent patch.

core/src/main/java/org/ldaptive/handler/RecursiveResultHandler.java
dfish3r: The entry gets mutated below, you need a copy in order to change the entry coming from a search operation.

core/src/main/java/org/ldaptive/LdapEntry.java
serac: Given the key importance of the flag in implementing immutability, I concede it may merit special treatment w/r/t under concurrency. That said, more analysis and test coverage in the future may be warranted.

core/src/main/java/org/ldaptive/auth/AuthenticationResponse.java
serac: Sounds good.

PR #264, opened 2024-07-12
additional logging for eDirectory response handler

mmoayyed:
The current response handler for eDirectory has no log statements, which makes it difficult to determine how it calculates the final account state.


dfish3r:
I took this PR and applied your logging changes to some other auth response handlers for consistency.
See c7a0e99c8ccc5dc7f5415cb26c159af58551b299

Thank you for the patch!

mmoayyed:
Thanks much!

PR #266, opened 2024-07-16
Remove serializable interface from JAAS principals.

dfish3r:
LdapPrincipal and LdapDnPrincipal both contain LdapEntry which is not serializable. Attempts to serialize these objects will throw.
Make LdapGroup immutable.


serac:
Looks good. I doubt anyone will notice the change in practice.

PR #274, opened 2024-11-14
Improve flexibility of prune strategy.

dfish3r:
PruneStrategy encapsulates the entire configuration for pruning, as such it can be difficult to implement complex pruning policies. Previously the connection pool itself restricted pruning such that connection could not be pruned below the minimum pool size. With this patch that logic is moved into the prune strategy allowing any and all available connections to be pruned. As a consequence, the connection pool will now attempt to grow back to its minimum size after pruning is complete. IdlePruneStrategy has been updated to support both a max connection age and a priority. Any connection exceeding the max age duration is removed from the pool. In conjuction with the priority setting, max age cannot also be applied to aggresively prune connections that are less desirable. These settings can provide a guarantee that some or all connections will be eventually recycled when the pool is idle.

### DIFF
--- a/core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
+++ b/core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
@@ -4,6 +4,7 @@ package org.ldaptive.pool;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -316,7 +317,6 @@ public abstract class AbstractConnectionPool extends AbstractFreezable implement
    *
    * @return  activator
    */
-  @Override
   public ConnectionActivator getActivator()
   {
     return activator;
@@ -328,7 +328,6 @@ public abstract class AbstractConnectionPool extends AbstractFreezable implement
    *
    * @param  a  activator
    */
-  @Override
   public void setActivator(final ConnectionActivator a)
   {
     assertMutable();
@@ -342,7 +341,6 @@ public abstract class AbstractConnectionPool extends AbstractFreezable implement
    *
    * @return  passivator
    */
-  @Override
   public ConnectionPassivator getPassivator()
   {
     return passivator;
@@ -354,7 +352,6 @@ public abstract class AbstractConnectionPool extends AbstractFreezable implement
    *
    * @param  p  passivator
    */
-  @Override
   public void setPassivator(final ConnectionPassivator p)
   {
     assertMutable();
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
```diff
@@ -1061,56 +1063,30 @@ public void prune()
...
-            callables.add(() -> {
-              if (pruneStrategy.apply(pc)) {
-                logger.trace("prune approved on {} with {} for {}", pc, pruneStrategy, AbstractConnectionPool.this);
-                return pc;
-              }
+        int numConnPruned = 0;
+        final int size = pruneStrategy.getPrunePredicates().size();
+        logger.trace("prune strategy has {} predicates", size);
+        // prune all available connections for each predicate supplied by the prune strategy
+        for (int i = 0; i < size; i++) {
```
serac, line 1070: This outer loop seems unnecessary. As I noted elsewhere I don't understand why you need to supply the index into the predicates list if they're evaluated in order. Seems like you could just iterate over them with an iterator as you do in the inner loop here.

serac (changes requested):
This is an improvement. I still feel that you can improve the API by internalizing the predicates and offering a simple consumer-based strategy:

    interface PruneStrategy implements Consumer<ConnectionPool>

You can still implement the predicates and implement the ordering over them in a base class. You need to think about how to protect calls to remove pool members, but a package-scoped method seems feasible.

core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
dfish3r: Each predicate must be invoked for each connection in the pool, in order.
The outer loop iterates over the predicates, the inner loop iterates over the pool.

core/src/main/java/org/ldaptive/pool/PruneStrategy.java
```diff
@@ -2,14 +2,19 @@
...
- * Provides an interface for pruning connections from the pool.
+ * Provides an interface for pruning connections from the pool. The list of predicates provided by the strategy are used
+ * against the entire pool in order. So that connections pruned by the first predicate are no longer available when the
+ * second predicate is used. In this fashion, a prune strategy can order any number of predicates, each of which can
+ * inspect the pool and make a determination which to prune a connection.
  *
  * @author  Middleware Services
  */
-public interface PruneStrategy extends Function<PooledConnectionProxy, Boolean>
+public interface PruneStrategy extends BiPredicate<PooledConnectionProxy, Integer>
```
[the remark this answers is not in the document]
dfish3r: You're right, I was fighting too hard to make `PruneStrategy` a functional interface. It only needs to supply the list of predicates. Will fix.

core/src/main/java/org/ldaptive/pool/PruneStrategy.java
```diff
@@ -27,4 +32,13 @@ public interface PruneStrategy extends Function<PooledConnectionProxy, Boolean>
...
   Duration getPrunePeriod();
+
+
+  /**
+   * Returns the predicates used by this prune strategy. Each predicate will be invoked for every connection to be
+   * pruned.
+   *
+   * @return  list of prune predicates
+   */
+  List<Predicate<PooledConnectionProxy>> getPrunePredicates();
```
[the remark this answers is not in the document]
dfish3r: Fixed.

dfish3r:
> This is an improvement. I still feel that you can improve the API by internalizing the predicates and offering a simple consumer-based strategy:
> 
> ```
> interface PruneStrategy implements Consumer<ConnectionPool>
> ```
> 
> You can still implement the predicates and implement the ordering over them in a base class. You need to think about how to protect calls to remove pool members, but a package-scoped method seems feasible.

I can see getting to 
```
interface PruneStrategy extends Function<Iterator<PooledConnectionProxy>, Integer>
```
where the function returns the number of connections pruned.

Passing the entire `ConnectionPool` introduces a lot of complexity for very little value. Ultimately, do you want people implementing an interface or extending a base class? Because the implementation people will care about is in `getPruneConditions`.  And where ever that ends up living is where you should expect people to customize the behavior.

core/src/main/java/org/ldaptive/pool/AbstractConnectionPool.java
serac: I see, thanks

core/src/main/java/org/ldaptive/pool/PruneStrategy.java
```diff
@@ -2,14 +2,19 @@
...
- * Provides an interface for pruning connections from the pool.
+ * Provides an interface for pruning connections from the pool. The list of predicates provided by the strategy are used
+ * against the entire pool in order. So that connections pruned by the first predicate are no longer available when the
+ * second predicate is used. In this fashion, a prune strategy can order any number of predicates, each of which can
+ * inspect the pool and make a determination whether to prune a connection.
  *
  * @author  Middleware Services
  */
-public interface PruneStrategy extends Function<PooledConnectionProxy, Boolean>
+public interface PruneStrategy extends Consumer<Supplier<Iterator<PooledConnectionProxy>>>
```
serac, line 17: At first glance this looks complex, but upon review of the abstract class it seems pretty natural: give me some connections to iterate over and remove the ones I don't like via `iterator#remove`. I think hiding the predicates from the public API is a notable improvement in clarity.

core/src/main/java/org/ldaptive/pool/PruneStrategy.java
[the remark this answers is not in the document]
serac: The new `Consumer`-based contact is much clearer.

serac:
Looks good. You probably need to tweak the commit message to reflect the API changes we worked through during review.

PR #275, opened 2024-11-22
Add LdapEntry#mapAttribute methods.

dfish3r:
These methods provide a convenient null-safe way to map the properties of an LdapAttribute to other values. This can reduce the amount of boilerplate code that retrieves an attribute, performs a null check and then performs an operation on the attribute. Also adds LdapEntry#hasAttribute convenience method.

### DIFF
--- a/core/src/main/java/org/ldaptive/LdapAttribute.java
+++ b/core/src/main/java/org/ldaptive/LdapAttribute.java
@@ -238,7 +238,7 @@ public class LdapAttribute extends AbstractFreezable
   public Collection<byte[]> getBinaryValues()
   {
     if (attributeValues.isEmpty()) {
-      return Collections.emptySet();
+      return Collections.emptyList();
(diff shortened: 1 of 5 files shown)


serac (approved):
I think this is a good improvement, though I would have liked to see you go further and expose null-safe value getters directly on `LdapEntry`:

- `public String getStringAttributeValue()`
- `public Collection<String> getStringAttributeValues()`
- `public byte[] getBinaryAttributeValue()`
- `public Collection<byte[]> getBinaryAttributeValues()`

It makes sense for these to delegate to your new `mapAttribute` method with the corresponding getter function already present on `LdapAttribute`. Seems like a worthwhile duplication for pattern that seems very common: search and pull data out of a result set by attribute name.

serac (approved):
Looks good. I appreciate your responding to my feedback.

PR #277, opened 2024-12-30
Avoid excessive memory consumption in DERParser.

dfish3r:
Because DERParser path permutations grow exponentially, a relatively short encoding can cause an OOM event. Restrict the DERParser path depth to 16, which seems more than adequate. Remove DefaultDERBuffer constructor that accepts an int, it isn't used and could potentially be an avenue for excessive memory use. Add DERBuffer#positionAndLimit to work around the subtleties of using both #position and #limit with different DERBuffer implementations. Add unit test for DERParser and ResponseParser.


core/src/main/java/org/ldaptive/asn1/DERParser.java
```diff
@@ -20,6 +20,12 @@
 public class DERParser
 {
 
+  /** Maximum supported path depth. */
+  private static final int MAX_PATH_DEPTH = 16;
+
+  /** Maximum number of tag permutations allowed. */
+  private static final int MAX_PERMUTATIONS = (int) Math.pow(2, MAX_PATH_DEPTH);
```
serac, line 27: I'm curious if this is strictly defensive.

serac (approved):
Looks good.

core/src/main/java/org/ldaptive/asn1/DERParser.java
dfish3r: Yes, we don't see BER path depths that long in practice.

PR #278, opened 2025-01-06
Support both errors and warnings in AccountState

dfish3r:
This PR provides (2) commits. (don't squash)

1. Change the cardinality of `AccountState` such that both errors and warnings can be set at the same time. While most systems generally don't support reporting on both errors and warnings, the ppolicy control does allow setting both.
2. Update `PasswordPolicyControl` to align with the definitions in the RFC by allowing both a single warning and an error to be set.

### DIFF
--- a/core/src/main/java/org/ldaptive/auth/ext/PasswordPolicyAccountState.java
+++ b/core/src/main/java/org/ldaptive/auth/ext/PasswordPolicyAccountState.java
@@ -6,7 +6,8 @@ import org.ldaptive.auth.AccountState;
 import org.ldaptive.control.PasswordPolicyControl;
 
 /**
- * Represents the state of an account as described by a password policy control.
+ * Represents the state of an account as described by a password policy control. The {@link PasswordPolicyControl}
+ * supports a single warning and/or a single error.
  *
  * @author  Middleware Services
  */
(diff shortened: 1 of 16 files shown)


core/src/main/java/org/ldaptive/auth/ext/PasswordPolicyAccountState.java
```diff
@@ -42,6 +53,32 @@ public PasswordPolicyAccountState(final PasswordPolicyControl.Error error)
...
+   */
+  public PasswordPolicyAccountState(final ZonedDateTime exp, final PasswordPolicyControl.Error error)
+  {
+    super(new AccountState.DefaultWarning(exp, -1), error);
+    ppError = error;
+  }
+
+
+  /**
+   * Creates a new password policy account state.
```
serac, line 70: It might be helpful to add some text about when each constructor form should be used. I can infer that this one is for _both_ an error and warning condition, but some text may be helpful.

serac (approved):
Looks good

core/src/main/java/org/ldaptive/auth/ext/PasswordPolicyAccountState.java
dfish3r: Fixed.

